PR #14, opened 2019-03-11
Add yarn lerna script to package.json

IreneStr:
This change makes it possible to use 'yarn lerna [whatever]', which means lerna doesn't have to be installed locally in order to use it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     ]
   },
   "scripts": {
+    "lerna": "lerna",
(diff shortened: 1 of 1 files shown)


IreneStr:
Closing, because not relevant anymore.

atimmer:
Not necessary because of this nice yarn feature: https://yarnpkg.com/en/docs/cli/#locally-installed-clis

PR #52, opened 2019-03-13
Ensure that yarn test in yoastseo installs the premium-configuration

### LINKED ISSUE #16: Add a message to yarn test if the premium-configuration is not installed
## Summary
The test suites fail if `yoastseo` does not have the `premium-configuration`. This can be fixed manually by running `grunt get-premium-configuration` in the `packages/yoastseo` directory. However, we want to give feedback to the user that this needs to happen.

## Explanation
Ensure that `yarn lint` gives an error if the `premium-configuration` is not installed. This allows the user to fix the test. The solution that fixes the problem is running `grunt get-premium-configuration` in the `packages/yoastseo` directory.

bintzandt:
As the premium-configuration is a prerequisite for the test, we should ensure that it is always installed and up-to-date before running the actual test.

This PR simply adds the `install-ci` command to the `yarn test` of yoastseo. This ensures that the premium-configuration is updated or installed if it is not already.

Fixes #16.


packages/yoastseo/package.json
```diff
@@ -18,7 +18,7 @@
     "start": "npm start --prefix examples/webpack",
     "build": "grunt publish",
     "lint": "grunt check",
-    "test": "jest --runInBand",
+    "test": "yarn install-ci && jest --runInBand",
```
diedexx: https://docs.npmjs.com/misc/scripts
You can use the `"pretest"` script here, which I think looks a bit better, as install-ci is not directly related to testing the code.

diedexx:
Acc 👍

Issue #46, opened 2019-01-18
Create a FacebookTextContainer component


andizer:
The `FacebookTextContainer` should contain the following components: 
* `FacebookSiteName` (Yoast/yoast-components#807) (and `FacebookAuthorName` (#42))
* `FacebookTitle` (#40) 
* `FacebookDescription` (#41)

Depending on the image type the container should have the following dimensions:
* Landscape: 
   * width: 500px. Height depends on the title and description, this needs a **decision**. It is probably 46px.
* Square mode: 
   * width: 318px and height: 136px
* Portait:
   * width: 318px and height: 215px

## Styling
The component should have the following CSS props:
```css
   background-color: #F2F3F5;
    border-bottom: 1px solid #dddfe2;
    margin: 0;
    padding: 10px 12px;
    position: relative;
```

The portrait/square modus will have the following rules:
```css
    border-top: 1px solid #dddfe2;
```

### Inner container
The component should have a 'inner container'.

When the Preview is in portrait or square mode the css the inner container has to following props:
```css
    display: flex;
    flex-direction: column;
    justify-content: center;
    height: 136px;
    font-size: 12px;
   margin: 0;
```

When its landscape the inner container has the following CSS props:
```css
    max-height: 190px;
```

### FacebookSiteName and the FacebookAuthorName
The `FacebookSiteName` and the `FacebookAuthorName` components must be wrapped in a div with the following CSS rules:
```css
   color: #606770;
    flex-shrink: 0;
    font-size: 12px;
    line-height: 16px;
    overflow: hidden;
    padding: 0;
    text-overflow: ellipsis;
    text-transform: uppercase;
    white-space: nowrap;
```

When the view is landscape it should have the following CSS specs:
```css
    position: relative;
```

For the square/portrait it should have:
```
   position: static;
```

#### Sidenotes about the length of the title and description
There is something vague about the title. Facebook has no description about the max length. When the title is long it will be shown on multiple lines (mostly 2 lines). When this is the case the description will be (probably) shortened. 

When the title is short (1 line) in most cases the description will be longer. There is no documentation about this behaviour (it's my own observation)

#### Wrapping in a div
The `FacebookTitle` and `FacebookDescription` components should be wrapped in a div. This `div` should have the following CSS specs when in portrait mode:
```css
    max-height: 46px;
    overflow: hidden;
```

##  How its been build in Facebook:
```html
<container>
     <subcontainer>
         <div>Sitenaam | author</div>
         <textcontainer>
[...]

PR #54, opened 2019-03-13
Create FacebookDescription component

### LINKED ISSUE #41: Create a FacebookDescription component
The `FacebookDescription` is shown as a div in the Facebook preview:
yoast_-_home

## Properties
* description

## Styling
The component should have the following CSS props:
```css
    color: #606770;
    font-size: 14px;
    line-height: 20px;
    word-break: break-word;
    max-height: 80px;
    overflow: hidden;
    text-overflow: ellipsis;
    white-space: normal;
    margin-top: 3px;
```

Parent https://github.com/Yoast/javascript/issues/46

igorschoester:
## Summary

* Adds a FacebookDescription component in SocialPreview\Facebook

* Requires a description property. The parent should take care of the event we do not have a description and not render this component (if this happens at all - e.g. we have a fallback).
* ~Use a fallback string for when the description is empty.~ No longer using a fallback here. It should be a simple component. The place to do this should probably be the Redux container.
* The stripping of the HTML is also no longer done here. Because if that results in an empty string we would want the fallback to be displayed. So this needs to be done in combination with the above.
* Add the FacebookDescription to the FacebookPreview directly for now (should be through FacebookTextContainer later).

## Test instructions

* Check the tests.
* Start the stand-alone.
* Go to the `FacebookPreview` page by clicking on the button in the top right.
* Check if they look alright. Mind you that the text container is not made, which will implement the sizes.
* Pay extra attention to the truncation in different browsers! IE, Firefox and Chrome.

Fixes #41

### DIFF
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
@@ -24,7 +24,9 @@ const FacebookDescriptionWrapper = styled.p`
  */
 const FacebookDescription = ( props ) => {
 	let description = string.stripHTMLTags( props.description );
-	if ( description.length === 0 ) description = "Modify your Facebook description by editing it right here";
+	if ( description.length === 0 ) {
+		description = "Modify your Facebook description by editing it right here";
+	}
 
 	return (
 		<FacebookDescriptionWrapper>
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
@@ -11,8 +11,9 @@ const FacebookDescriptionWrapper = styled.p`
 	max-height: 80px;
 	overflow: hidden;
 	text-overflow: ellipsis;
-	white-space: normal;
+	white-space: nowrap;
 	margin-top: 3px;
+	width: 100%;
 `;
 
 /**
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
@@ -23,9 +23,12 @@ const FacebookDescriptionWrapper = styled.p`
  * @returns {React.Element} The rendered element.
  */
 const FacebookDescription = ( props ) => {
+	let description = string.stripHTMLTags( props.description );
+	if ( description.length === 0 ) description = "Modify your Facebook description by editing it right here";
+
 	return (
 		<FacebookDescriptionWrapper>
-			{ string.stripHTMLTags( props.description ) }
+			{ description }
 		</FacebookDescriptionWrapper>
 	);
 };
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
@@ -1,0 +1,41 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+import { string } from "yoastseo";
+
+const FacebookDescriptionWrapper = styled.p`
+    color: #606770;
+    font-size: 14px;
+    line-height: 20px;
+    word-break: break-word;
+    max-height: 80px;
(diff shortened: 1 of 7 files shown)


Xyfi:
CR 👍

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+
+const FacebookDescriptionWrapper = styled.p`
+    color: #606770;
```
afercia: Ideally, all colors should be set in the color palette. Not sure what to do in this case, as it's a single-use color.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
```diff
@@ -0,0 +1,36 @@
...
+    font-size: 14px;
+    line-height: 20px;
+    word-break: break-word;
+    max-height: 80px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: normal;
+    margin-top: 3px;
+`;
```
afercia: All the indentation in this ruleset is made with spaces. Should be changed to tab.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
```
afercia: I'd add the "external dependencies" comment.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookPreview.js
```diff
@@ -18,17 +19,20 @@ const FacebookPreview = ( props ) => {
...
 			<FacebookImage src={ props.src } alt={ props.alt } />
 			<FacebookSiteName siteName={ props.siteName } />
+			<FacebookDescription description={ props.description } />
 		</Fragment>
 	);
 };
 
 FacebookPreview.propTypes = {
 	siteName: PropTypes.string.isRequired,
+	description: PropTypes.string,
```
afercia: description is required in `FacebookDescription` but it's not required here?

afercia:
Acceptance 🚧 

Pushed a few minor coding standards. However, there's a problem.

`text-overflow: ellipsis;` and `white-space: normal;` don't work together. As the text can wrap in new lines, the ellipsis won't be displayed. What is the desired behavior?
- if we want ellipsis, I guess it can't be done with CSS and should be done via JS
- if we don't want ellipsis, `text-overflow: ellipsis;` should be removed

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookDescription.js
igorschoester: I reckon this is an exception to that rule because of this not being a Yoast color.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookPreview.js
igorschoester: Sorry, forgot to update the PR description.

The `FacebookDescription` should be a simple component that does require the prop. While the parent (`FacebookTextContainer` in the future) should then handle to not render the description if not needed.
In the `FacebookTextContainer` we'd probably want this check.

igorschoester:
Facebook does things a bit differently. It uses `line-clamp`, which is in a working draft. IE and Firefox do not support this.
Facebook truncates the text themselves in Firefox.
Instead of trying to do it exactly as Facebook I chose to do it the easier way: Using `white-space: nowrap;` instead.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookPreview.js
johannadevos: @igorschoester I think you should add this information (i.e., that we want this check) to the FacebookTextContainer issue (#46).

johannadevos:
CR 👍

johannadevos:
Acceptance 👍 

As discussed, @igorschoester will add the things that still need to be done in the future (i.e., using the Redux container for empty descriptions, stripping HTML) to the epics.

PR #55, opened 2019-03-13
Implement functionality to determine a unique stem for German words

manuelaugustin:
## Summary

* [non-user-facing] Creates unique stems for German words.

* The functionality introduced here isn't used to create German word forms, but intended to be used for internal linking.
* For words that are on exception lists in multiple categories, we cannot currently determine the correct stem. E.g. "schwamm" could either be the noun "Schwamm" ("sponge") or the past tense form "schwamm" of the verb "schwimmen" (to swim).

## Test instructions

* Run `yarn test` to confirm that the coverage of the added functionality is 100%.
* Add additional German words to the spec and check whether they get stemmed correctly. Specifically, when you add different forms of a same word, they should receive the same stem.

Fixes https://github.com/Yoast/YoastSEO.js/issues/2203


packages/yoastseo/spec/morphology/german/determineStemSpec.js
```diff
@@ -0,0 +1,40 @@
+import { determineStem } from "../../../src/morphology/german/determineStem";
+import getMorphologyData from "../../specHelpers/getMorphologyData";
+
+
+const morphologyDataDE = getMorphologyData( "de" ).de;
+
+const nounsToStem = [
```
maartenleenders:
```suggestion
const wordsToStem = [
```

packages/yoastseo/spec/morphology/german/determineStemSpec.js
```diff
@@ -0,0 +1,40 @@
...
+	[ "überfraß", "überfress" ],
+	// Verbs: strongVerbs, not treated as prefixed
+	[ "berg", "berg" ],
+	// Participles
+	[ "heraufgeholt", "heraufhol" ],
+];
+
+describe( "Test for determining stems for German words", () => {
+	it( "creates stems for German words", () => {
+		nounsToStem.forEach( wordToStem => expect( determineStem( wordToStem[ 0 ], morphologyDataDE ) ).toBe( wordToStem[ 1 ] ) );
```
maartenleenders:
```suggestion
		wordsToStem.forEach( wordToStem => expect( determineStem( wordToStem[ 0 ], morphologyDataDE ) ).toBe( wordToStem[ 1 ] ) );
```

packages/yoastseo/src/morphology/german/determineStem.js
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Returns a stem for a word that appears on the adjective exception lists.
+ *
+ * @param {Object}  morphologyDataAdjectives    The German morphology data for adjectives.
+ * @param {string}  stemmedWord                 The stem to check.
+ *
+ * @returns {string|null} The stemmed word or null if none was found.
+ */
+const findStemOnAdjectiveExceptionList = function( morphologyDataAdjectives, stemmedWord ) {
+	const adjectiveExceptionClasses = {
```
maartenleenders: Rather than picking certain fields from the adjectives object, it might be wise to already put these exception classes in a subfield of adjectives (e.g. adjectives.exceptions). This makes it easier to extend the exceptionClasses (i.e. you don't have to edit the classes in the config and here).

maartenleenders: If that's not feasible for now, create an array with hard coded keys `[ "elStemChange", "erStemChangeClass1",`... etc, and loop over that, rather than copying the key-value pairs to another object, and looping over its `Object.keys()`.

packages/yoastseo/src/morphology/german/determineStem.js
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * @param {Object}  morphologyDataVerbs The German morphology data for verbs.
+ * @param {string}  stemmedWord         The stem to check.
+ *
+ * @returns {string|null} The stemmed word or null if none was found.
+ */
+const findStemOnVerbExceptionList = function( morphologyDataVerbs, stemmedWord ) {
+	let wordToCheck = stemmedWord;
+	const strongVerbStems = morphologyDataVerbs.strongVerbs.stems;
+	const prefixes = morphologyDataVerbs.verbPrefixes;
```
maartenleenders: Optionally, refactor the verbPrefixes field in the morphology, because it seems to be a combination of other prefix-related fields.

packages/yoastseo/src/morphology/german/determineStem.js
```diff
@@ -0,0 +1,125 @@
...
+
+	let matchedPrefix = prefixes.find( prefix => stemmedWord.startsWith( prefix ) );
+
+	if ( matchedPrefix ) {
+		const wordWithoutPrefix = wordToCheck.slice( matchedPrefix.length, wordToCheck.length );
+
+		// At least 3 characters so that e.g. "be" is not found in the stem "berg".
```
maartenleenders: Perhaps expand this comment slightly to explain the rationale behind the length of 3 characters.

packages/yoastseo/src/morphology/german/determineStem.js
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+
+	for ( const strongVerbParadigm of strongVerbStems ) {
+		let stems = strongVerbParadigm.stems;
+		stems = flatten( Object.values( stems ) );
+
+		if ( stems.includes( wordToCheck ) ) {
+			if ( matchedPrefix ) {
+				return matchedPrefix + stems[ 0 ];
```
maartenleenders: Because the order of the array might not be guaranteed when using `Object.values`, the word returned by `stems[ 0 ]` is not either. Could this be problematic?

packages/yoastseo/src/morphology/german/getForms.js
```diff
@@ -124,7 +124,7 @@ export function getForms( word, morphologyData ) {
 		// Add the original word as a safeguard.
 		exceptions.push( word );
 
-		return { forms: unique( exceptions ), stem: stemmedWord };
+		return unique( exceptions );
```
maartenleenders: JSDoc return value not updated

maartenleenders:
CR 🚧 
Some minor suggestions / considerations

manuelaugustin:
Thanks for the feedback! I've implemented all points as suggested.

nataliashitova:
CR 👍 
@manuelaugustin Please fix the JSDoc as suggested by Maarten https://github.com/Yoast/javascript/pull/55/files#r265648752

manuelaugustin:
Oh right, I had overlooked that one. Fixed now!

hansjovis:
Acceptance test: ✅ .

All unit tests pass. I added a few more tests to check the German stemming functionality.
One line of `detectAndStemRegularParticiple.js` was uncovered, so I added a stem-test to cover this line as well.

PR #57, opened 2019-03-13
Introduces the @yoast/components package

### LINKED ISSUE #30: Create the components package
## Summary

Creates a components package: `packages/components` that will be accessible from `@yoast/components`.  

## Explanation

What we have to do is:
* [x] Create a new directory: `packages/components`
* [x] Moving the contents of `packages/yoast-components/forms/` to this directory
* [x] Make all components from the old `forms` directory is accessible/referenced in `packages/yoast-components/index.js`
* [x] Moves the contents of `packages/yoast-components/composites/basic/` to the root of `packages/components`
* [x] Change all references from '/basic' to the new locations in `packages/yoast-components/index.js`
* [x] Makes sure all loaded components are loaded from the new locations.

andizer:
Fixes #30 

Makes sure `yarn lint` and `yarn test` will have the same output as on the `develop` branch


bintzandt:
All dependencies need to be added to `package.json` (we can't rely on having the same context everywhere since these packages can be used separately).
Furthermore, all references to `yoast-components` should be to the package (which also needs to be added to `package.json`) thus `../yoast-components/bla/bla` needs to be changed to `yoast-components`.

packages/components/ErrorBoundary.js
```diff
@@ -6,7 +6,7 @@ import { __ } from "@wordpress/i18n";
 import { speak as a11ySpeak } from "@wordpress/a11y";
 
 /* Internal dependencies */
-import colors from "../../style-guide/colors";
+import colors from "../yoast-components/style-guide/colors";
```
atimmer: We should have no dependencies on `yoast-components` in `@yoast/components`. So that means we should probably move `colors` to `@yoast/components`.

packages/yoast-components/index.js
```diff
@@ -27,7 +26,7 @@ import { default as LinkSuggestions } from "./composites/LinkSuggestions/LinkSug
 // Composites/KeywordSuggestions imports.
 import { default as KeywordSuggestions } from "./composites/KeywordSuggestions/KeywordSuggestions";
 // Composites/basic imports.
-import { default as Loader } from "./composites/basic/Loader";
+import { default as Loader } from "../components/Loader";
```
diedexx: Can this be written as "@yoast/components/Loader`?

bintzandt: This is something that needs to be changed.

bintzandt:
Also, the build is still failing.

packages/yoast-components/composites/OnboardingWizard/onboarding-wizard.scss
```diff
@@ -1,4 +1,4 @@
-@import "../../style-guide/colors";
+@import "../../../components/style-guide/colors";
```
diedexx: This could be written with `~`

packages/yoast-components/css/all.scss
```diff
@@ -1,4 +1,4 @@
 @import "accessibility";
 @import "../composites/OnboardingWizard/onboarding-wizard";
-@import "../composites/basic/loader";
+@import "../../components/loader";
```
diedexx: This cuold be written with `~`

PR #60, opened 2019-03-13
Yoastseo/2196 rewrite analyze to use structured tree

hansjovis:
## Summary

* _[not-user-facing]_ Rewrites the SEO and readability analysis such that it uses both the original assessors as well the new tree assessors. The results of both are combined and their scores aggregated.

* The `wrapTryCatchAroundAction` that handles the errors thrown in the web worker is now partly asynchronous, to make it work with the new asynchronous nature of the analysis.

## Test instructions

* Start the example.
* Type in a (HTML) text.
* Check if the analysis works as expected, and does show an error in the console.
    * A few errors are known already: #38 and #61.
* Run the tests, check if they pass.

Fixes https://github.com/Yoast/YoastSEO.js/issues/2196
Fixes https://github.com/Yoast/YoastSEO.js/issues/2205


hansjovis:
I still need to add error handling around the building of the tree.
If it fails, we can still analyze the contents with the old assessor and return the results on the assessments that have not been migrated yet.
We should show an assessment result error (gray bullet) for those assessments that use the tree builder when it fails.

manuelaugustin:
CR 🚧

All the code itself looks good to me!

I just had some small suggestions to make the formatting of single line comments more consistent.

Also there is some newly introduced code that's not covered by unit tests. These things are very minor, but see if you can maybe at tests to cover them:
In `AnalysisWebWorker.js`
Line 450 - else path of if-statement not covered
Line 759 - catch statement isn't covered

manuelaugustin:
Acceptance 🚧

I tried to induce an error to make the building of the tree fail (using the example from https://github.com/Yoast/javascript/issues/38). Rather than being caught, the error caused the dev tool to crash. I think there's some further investigation required here.

hansjovis:
> Acceptance 🚧
> 
> I tried to induce an error to make the building of the tree fail (using the example from #38). Rather than being caught, the error caused the dev tool to crash. I think there's some further investigation required here.

I identified the issue. The example has a toggle to build the tree for visualization purposes. This building did not have error handling yet. I made an issue and a PR that fixes this: #130 .

manuelaugustin:
Acceptance 👍

PR #63, opened 2019-03-14
Create FacebookAuthorName component

### LINKED ISSUE #42: Create a FacebookAuthorName component
The `FacebookAuthorName` should be placed within the `FacebookSiteName` component (https://github.com/Yoast/yoast-components/issues/807)
yoast_-_home

When the name is present it should start with a separator (`|`). The name itself should be styled as a link (but it is not a link).

## Styling
The styling for the separator (`span`) is 
```css
    padding-left: 5px;
    padding-right: 5px;
}
```

The styling for the name (link) will be:
```css
    color: #3b5998;
    position: relative;
    text-decoration: none;
    cursor: pointer;
    z-index: 5;
```

Needs https://github.com/Yoast/yoast-components/issues/807

Parent #46

igorschoester:
## Summary

* Adds a FacebookAuthorName component.
* Adds a FacebookSiteAndAuthorNames component.

* Instead of placing the `FacebookAuthorName` within the `FacebookSiteName` I created a `FacebookSiteAndAuthorNames` to hold them both.
* Changed `FacebookSiteName`'s HTML element to span because it should be inline.
* Adapted `FacebookPreview` to use `FacebookSiteAndAuthorContainer` instead of `FacebookSiteName`.

## Test instructions

* Check the tests.
* Start the stand-alone.
* Go to the FacebookPreview page by clicking on the button in the top right.
* Check if they look according to spec (see issue or better yet Facebook itself).

## UI changes

Fixes #42

### DIFF
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
@@ -1,0 +1,33 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+
+const FacebookAuthorNameWrapper = styled.span`
+    color: #3b5998;
+    font-size: 12px;
+    line-height: 11px;
+    position: relative;
+    text-decoration: none;
+    cursor: auto;
+`;
+
+/**
+ * Renders a FacebookAuthorName component.
+ *
+ * @param {object} props The props.
+ *
+ * @returns {React.Element} The rendered element.
+ */
+const FacebookAuthorName = ( props ) => {
+	return (
+		<FacebookAuthorNameWrapper>
+			{ props.authorName }
+		</FacebookAuthorNameWrapper>
+	);
+};
+
+FacebookAuthorName.propTypes = {
+	authorName: PropTypes.string.isRequired,
+};
+
+export default FacebookAuthorName;
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
@@ -4,12 +4,12 @@ import styled from "styled-components";
 import PropTypes from "prop-types";
 
 const FacebookAuthorNameWrapper = styled.span`
-    color: #3b5998;
-    font-size: 12px;
-    line-height: 11px;
-    position: relative;
-    text-decoration: none;
-    cursor: auto;
+	color: #3b5998;
+	font-size: 12px;
+	line-height: 11px;
+	position: relative;
+	text-decoration: none;
+	cursor: auto;
 `;
 
 /**
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
@@ -7,9 +7,6 @@ const FacebookAuthorNameWrapper = styled.span`
 	color: #3b5998;
 	font-size: 12px;
 	line-height: 11px;
-	position: relative;
-	text-decoration: none;
-	cursor: auto;
 `;
 
 /**
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
@@ -1,3 +1,4 @@
+/* External dependencies */
 import React from "react";
 import styled from "styled-components";
 import PropTypes from "prop-types";
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
@@ -1,8 +1,10 @@
+/* External dependencies */
 import React, { Fragment } from "react";
 import styled from "styled-components";
 import PropTypes from "prop-types";
 import { __ } from "@wordpress/i18n";
 
+/* Internal dependencies */
 import FacebookSiteName from "./FacebookSiteName";
 import FacebookAuthorName from "./FacebookAuthorName";
 
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
@@ -9,13 +9,13 @@ import FacebookSiteName from "./FacebookSiteName";
 import FacebookAuthorName from "./FacebookAuthorName";
 
 const FacebookSiteAndAuthorNamesWrapper = styled.p`
-    color: #606770;
-    font-size: 12px;
-    line-height: 11px;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    text-transform: uppercase;
-    white-space: nowrap;
+	color: #606770;
+	font-size: 12px;
+	line-height: 11px;
+	overflow: hidden;
+	text-overflow: ellipsis;
+	text-transform: uppercase;
+	white-space: nowrap;
 `;
 
 const FacebookSiteAndAuthorNamesSeparator = styled.span`
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
@@ -1,0 +1,75 @@
+import React, { Fragment } from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+import { __ } from "@wordpress/i18n";
+
+import FacebookSiteName from "./FacebookSiteName";
+import FacebookAuthorName from "./FacebookAuthorName";
+
+const FacebookSiteAndAuthorNamesWrapper = styled.p`
+    color: #606770;
+    font-size: 12px;
+    line-height: 11px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    text-transform: uppercase;
+    white-space: nowrap;
+`;
+
+const FacebookSiteAndAuthorNamesSeparator = styled.span`
+    padding-left: 5px;
+    padding-right: 5px;
+`;
+
+/**
+ * Renders a FacebookAuthorName component with accompanying elements.
+ *
+ * @param {string} authorName The author's name.
+ *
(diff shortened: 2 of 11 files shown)


packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
```
afercia: I'd add the external dependencies comment.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
```diff
@@ -0,0 +1,33 @@
...
+import styled from "styled-components";
+import PropTypes from "prop-types";
+
+const FacebookAuthorNameWrapper = styled.span`
+    color: #3b5998;
```
afercia: Ideally, colors should be in the color palette. Not sure in this case, as this is a single-use color.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
```diff
@@ -0,0 +1,33 @@
...
+    font-size: 12px;
+    line-height: 11px;
```
afercia: Any reason why the line-height is so small? Also, ideally it should be unitless. I'd go for `line-height: 1;`

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
```diff
@@ -0,0 +1,33 @@
...
+    position: relative;
```
afercia: position: relative; doesn't do anything here, it should be removed.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
```diff
@@ -0,0 +1,33 @@
...
+    text-decoration: none;
```
afercia: text-decoration: none; doesn't do anything: this is not a link and it's not underlined by default. It should be removed.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
```diff
@@ -0,0 +1,33 @@
...
+    cursor: auto;
```
afercia: cursor: auto; doesn't do anything: this is not a link, just remove it please.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
```diff
@@ -0,0 +1,33 @@
...
+`;
```
afercia: all the indentation in this ruleset is made with spaces, should be changed to tabs.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
```diff
@@ -0,0 +1,75 @@
+import React, { Fragment } from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+import { __ } from "@wordpress/i18n";
+
+import FacebookSiteName from "./FacebookSiteName";
+import FacebookAuthorName from "./FacebookAuthorName";
+
+const FacebookSiteAndAuthorNamesWrapper = styled.p`
+    color: #606770;
```
afercia: same as above about colors

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
```diff
@@ -0,0 +1,75 @@
...
+    font-size: 12px;
+    line-height: 11px;
```
afercia: same as above about line-height

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
```diff
@@ -0,0 +1,75 @@
...
+    overflow: hidden;
+    text-overflow: ellipsis;
+    text-transform: uppercase;
+    white-space: nowrap;
+`;
```
afercia: same as above: all this ruleset uses spaces for indentation, should be changed to tabs

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
```diff
@@ -0,0 +1,75 @@
...
+
+const FacebookSiteAndAuthorNamesSeparator = styled.span`
+    padding-left: 5px;
+    padding-right: 5px;
+`;
```
afercia: same as above: all this ruleset uses spaces for indentation, should be changed to tabs

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteName.js
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import styled from "styled-components";
 import PropTypes from "prop-types";
 
-const FacebookSiteNameWrapper = styled.p`
+const FacebookSiteNameWrapper = styled.span`
```
afercia: whit this change, some of the CSS rules should be removed.
text-overflow: ellipsis; doesn't work on inline elements
as a consequence, white-space: nowrap; should be removed as well (the on on the container will work)
re: the line-height and color, see previous considerations

afercia:
CR 🚧 

See comments, mainly CSS stuff.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
johannadevos: Done

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
```diff
@@ -0,0 +1,75 @@
+import React, { Fragment } from "react";
```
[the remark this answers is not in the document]
johannadevos: Done

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
johannadevos: @afercia What do you mean with "the color palette"?

afercia: @johannadevos a while ago we've built a color palette to control all the colors used in the components. Then in the components we should use the variables representing the colors. See 
https://github.com/Yoast/javascript/blob/develop/packages/yoast-components/style-guide/_colors.scss which then gets compiled in a [json file](https://github.com/Yoast/javascript/blob/develop/packages/yoast-components/style-guide/colors.json) for reuse in JavaScript. 
In this specific case I'd agree with @igorschoester: the Facebook and Twitter colors aren't part of our design so maybe we can make an exception.

johannadevos: @afercia Thanks for the explanation! @igorschoester and @IreneStr both agree that we should make an exception for Facebook and Twitter colours :)

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
johannadevos: The reason is that we're using the exact same CSS settings as Facebook, and they use `line-hight: 11px;`.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
johannadevos: The reason is that we're using the exact same CSS settings as Facebook, and they use `line-hight: 11px;`.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
johannadevos: Done

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
johannadevos: Done

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
johannadevos: Done

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteName.js
johannadevos: Done

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookAuthorName.js
johannadevos: Done

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
johannadevos: Done

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
johannadevos: Done

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
johannadevos: See above

IreneStr:
CR 👍

IreneStr:
Acceptance 👍 
I've created an issue to address some screenreader concerns: https://github.com/Yoast/javascript/issues/89

PR #64, opened 2019-03-18
Add a deploy stage for lerna publishing

### LINKED ISSUE #15: Setup auto-publishing with lerna from Travis
If we use `lerna publish` from people's local machine then they all need publishing rights on all the packages. It's better to do it with Travis. We can use `lerna publish from-git` for this purpose. When people want to release they do a `lerna version` and then Travis does the rest after pushing to git.

atimmer:
The workflow will be as follows:

* The team to release the monorepo pushes a new version using `yarn lerna-version` or `yarn lerna-version-rc`. 
* Travis will start building, either on `master` or the `release-yoast-seo/[version]` branch.
* Travis executes the correct deployment stage (either `deployment` or `rc-deployment`)
* Travis publishes using `yarn lerna-publish` or `yarn lerna-publish-rc`.

This is near impossible to test yourself, so I will be working with @manuelaugustin to test the RC version on the `release-yoast-seo/10.1` branch. So, after code review you may merge this PR.

## Technical choices

I've chosen to put the decisions whether to deploy in the build stages definition. This makes it easy to see which stage has been executed without having to dig through the logs of a job. It also doesn't spin up a machine when that is unnecessary.

Fixes #15.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,37 @@ node_js:
 matrix:
   allow_failures:
     - node_js: node
+stages:
+  - test
+  - name: 🚀 rc-deployment
+    if: branch =~ /^(release|release-yoast-seo)\/*/
+  - name: 🚀 deployment
+    if: branch = master
+jobs:
+  include:
+    - stage: 🚀 rc-deployment
+      node_js: lts/*
+      name: "Deploy to the npm rc channel"
+      script: skip
+      deploy:
+        skip_cleanup: true
+        provider: script
+        script: yarn lerna-publish-rc
+        on:
+          # Branch condition is in the stages configuration
+          all_branches: true
+    - stage: 🚀 deployment
+      node_js: lts/*
+      name: "Deploy to the npm latest channel"
+      script: skip
+      deploy:
+        skip_cleanup: true
+        provider: script
+        script: yarn lerna-publish
+        on:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -6,6 +6,37 @@ node_js:
 matrix:
   allow_failures:
     - node_js: node
+stages:
+  - test
+  - name: 🚀 rc-deployment
+    if: branch =~ /^(release|release-yoast-seo)\/*/
```
bintzandt: I have tested the regex using [regexr](https://regexr.com) and it does not seem to work as intended. Also, I think the following regex is more readable :)

```
/^(release(-yoast-seo)?)\/*/
```

atimmer: Seems like it works?

Screen Shot 2019-03-18 at 15 55 46

Also, I think `release|release-yoast-seo` is more readable because in your proposed regex you need to parse the regex yourself to see which branches it applies to.

atimmer: Oh, I guess we need to swap the two branch names.

bintzandt: Yeah, swapping helps :) Both the regexes are fine by me, if you prefer the other one: go ahead. Just make sure to swap the names 👍

atimmer:
I've changed the order.

PR #66, opened 2019-03-18
Move yoast components example app to apps directory

### LINKED ISSUE #39: Move yoast-components example app to `apps` directory
## Summary

Alongside the `packages` folder we are going to need an `apps` folder that contains JavaScript that we are not planning on releasing to `npm`, but that still needs to do a job.

## Explanation

The first app that is going to move to the `apps` folder is the yoast-components example app, situated in `yoast-components/app`. Currently the app is started using `yarn start` in the yoast-components folder. I would say that the new app lives in `apps/components` and can be started using `yarn start` there.

All the configuration required to start this app (stubby, webpack, etc) needs to move to this app folder. We need to configure lerna so it will make sure the dependencies of these apps are still put in `node_modules`.

bintzandt:
Fixes #39 

# Summary
Moves the yoast-components app directory to the root and add this directory to `lerna`. This allows us to create a seperate package for demonstrating the components that we have.


apps/components/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@yoast/example-component-app",
+  "version": "1.0.0",
+  "description": "This app demonstrates the possibilites of our components",
+  "main": "index.js",
+  "author": "Yoast",
+  "license": "GPL",
```
diedexx: Should this be `GPL-3.0`?

diedexx:
Does this mean we can drop the `webpack`, `stubby`, `react-hot-reloader` and so on dependencies from yoast-components? And also, can we drop the [`start` script from the `package.json` ](https://github.com/Yoast/javascript/blob/develop/packages/yoast-components/package.json#L8) (or even better, replace it with instructions on how to start the newly made app?)

bintzandt:
> Does this mean we can drop the `webpack`, `stubby`, `react-hot-reloader` and so on dependencies from yoast-components? And also, can we drop the [`start` script from the `package.json` ](https://github.com/Yoast/javascript/blob/develop/packages/yoast-components/package.json#L8) (or even better, replace it with instructions on how to start the newly made app?)

Yes. I think that we can do that. However, it might be better to a) First test that and b) Maybe create a new PR for that (although that might be a bit overkill idk).

apps/components/package.json
bintzandt: Anton said `GPL` but I noticed that @andizer also used `GPL-3.0` so maybe we should indeed all use the same :D

diedexx:
👍 
I've pushed one commit (ef94c3a534c741ab2a7195ff3d8ea675c7ecae5f)  to prevent the teams who are using this to loose their !@#$ and spend 2 hours debugging why the start command fails.
Acc 👍 _waiting for travis to succeed_.

PR #70, opened 2019-03-18
Fix parsing of (unintended) comments

### LINKED ISSUE #38: Generating tree structures for HTML with a comment inside a heading or paragraph fails.
Generating the tree structure fails for HTML that has a comment within a header or paragraph:
```html
<h1>Hello World!<!-- How are you? --></h1>
```
The comment gets added to the header as a formatting element, as it should be, but something goes wrong when converting the comment from an `Ignored` node to a `FormattingElement`.

It currently errors on the calculation of the start- and end position of the formatting elements.
Specifically on this line, where it fails to get the start tag.
https://github.com/Yoast/javascript/blob/47b317eb2722575a1a504d522d2ff533580141a8/packages/yoastseo/src/tree/builder/cleanup/calculateTextIndices.js#L116

**More importantly**, `parse5` recognizes ` How are you?` in the example below as a comment:
```html
<h1>Hello World!</ How are you?
```
Which leads to an error when typing in a closing tag in a pre-existing text.

maartenleenders:
## Summary
* _[not-user-facing]_ Parsing comments inside paragraphs and headings does not break the analysis anymore.

Apparently, the browser parses incomplete tags as comments, like this:
```html
<p>This is a paragraph </ with an incomplete closing tag.</p>
```
to this:
```html
<p>This is a paragraph <!-- with an incomplete closing tag.</p --></p>
```
This broke our tree parser, since it could not handle comments inside paragraphs or headings.

This PR fixes this issue by returning the formatting elements to parse5 when it tries to retrieve the comment node from the tree.

## Relevant technical choices
* We refactored the code to allow `Ignored` nodes as formatting in a heading or paragraph. Previously, we transformed these nodes to `FormattingElement`s first.
  * We chose this option to be able to keep track of the source code location and content.

## Testing instructions
* In one of the example texts, add `</` (be mindful of PHPStorm's autocomplete) in a heading and a paragraph, and see that it still works. Do the same with `<!-- bla bla -->`, and `<script>blabla</script>`. Verify that this breaks on develop but not on this branch.

Fixes #38


maartenleenders:
Acceptance 👍

PR #74, opened 2019-03-19
Create FacebookTitle component

### LINKED ISSUE #40: Create a FacebookTitle component
The `FacebookTitle` is the bold piece of the Facebook preview:
yoast_-_home

The element has been rendered as a `div` element on Facebook.

## Sidenotes about the length
There is something vague about the title. Facebook has no description about the max length. When the title is long it will be shown on multiple lines (mostly 2 lines). When this is the case the description will be (probably) shortened. 

When the title is short (1 line) in most cases the description will be longer. There is no documentation about this behaviour (it's my own observation)

## Properties
* title

## Styling
The component should have the following CSS props:
```css
    color: #1d2129;
    font-weight: 600;
    overflow: hidden;
    font-size: 16px;
    line-height: 20px;
    margin: 5px 0 0;
    max-height: 110px;
    word-wrap: break-word;
    letter-spacing: normal;
    text-overflow: ellipsis;
    white-space: normal;
    cursor: pointer; // It is a link on Facebook
```
Parent https://github.com/Yoast/javascript/issues/46

johannadevos:
## Summary

* [non-userfacing] Adds a FacebookTitle component in SocialPreviews/Facebook

* I created the FacebookTitle component.
* I added the FacebookTitle component to the FacebookPreview component.

## Test instructions

* Check out this branch.
* From the `packages/yoast-components` folder in the terminal, run `yarn` and `yarn start`.
* Run the unit tests (`FacebookTitleTest.js` and `FacebookPreviewTest.js`), they should not result in any errors.
* Start the stand-alone by navigating to `http://localhost:3333/` in the browser.
* Go to the `FacebookPreview` page by clicking on the button in the top right.
* Check if the titles look alright.

## UI changes

## Quality assurance

Fixes #40

### DIFF
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookTitle.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookTitle.js
@@ -3,16 +3,16 @@ import styled from "styled-components";
 import PropTypes from "prop-types";
 
 const FacebookTitleWrapper = styled.span`
+	display: block;
 	color: #1d2129;
 	font-weight: 600;
 	overflow: hidden;
 	font-size: 16px;
 	line-height: 20px;
 	margin: 5px 0 0;
-	max-height: 110px;
+	max-height: 100px;
 	word-wrap: break-word;
 	letter-spacing: normal;
-	text-overflow: ellipsis;
 	white-space: normal;
 	cursor: pointer;
 `;
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookTitle.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookTitle.js
@@ -1,0 +1,39 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+
+const FacebookTitleWrapper = styled.span`
+	color: #1d2129;
+	font-weight: 600;
+	overflow: hidden;
+	font-size: 16px;
+	line-height: 20px;
+	margin: 5px 0 0;
+	max-height: 110px;
+	word-wrap: break-word;
+	letter-spacing: normal;
+	text-overflow: ellipsis;
+	white-space: normal;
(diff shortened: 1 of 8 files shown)


packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookTitle.js
```diff
@@ -0,0 +1,39 @@
...
+	color: #1d2129;
+	font-weight: 600;
+	overflow: hidden;
+	font-size: 16px;
+	line-height: 20px;
+	margin: 5px 0 0;
+	max-height: 110px;
+	word-wrap: break-word;
+	letter-spacing: normal;
+	text-overflow: ellipsis;
```
afercia: `text-overflow: ellipsis;` doesn't work with `white-space: normal;` nor with inline elements (the span) and should be removed. I know it is in the instructions in the issue :) but this won't have any effect.

packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookTitle.js
```diff
@@ -0,0 +1,39 @@
...
+import PropTypes from "prop-types";
+
+const FacebookTitleWrapper = styled.span`
+	color: #1d2129;
+	font-weight: 600;
+	overflow: hidden;
+	font-size: 16px;
+	line-height: 20px;
+	margin: 5px 0 0;
+	max-height: 110px;
```
afercia: `max-height` (and consequently `overflow: hidden`) doesn't work on inline elements.
Also, given the nwe line-height is 20 pixels we should change this value to 20*5 lines = 100 pixels.

igorschoester:
Acceptance ✅ 

I do not think the use of a `span` here is correct. After asking Andrea, he mentioned `span > div` would be invalid. This could possibly be the case in the FacebookTextContainer (if this is a `div` and the text container is a `span`). Therefore, I do not think this is blocking a merge.
If we chose to alter it then, the style `display: block` should be removed here.

Issue #31, opened 2019-03-12
Abstract search-metadata-previews into package


diedexx:
## Summary
We need a new package for the search-metadata-previews: `@yoast/search-metadata-previews`.

## Explanation
The contents of these composites should be included in this new package:
- yoast-components/composites/SnippetEditor
- yoast-components/composites/SnippetPreview

yoast-components should still export these composites if they did before, so we don't break backwards compatibility.
All usages of the composites inside the javascript repository should be updated. They should use the new `@yoast/search-metadata-previews` package.

diedexx:
We've made progress on the `add-search-hmetadata-preview-package` (yes with the typo 😓).

We cannot continue until the [`@yoast/components` package](https://github.com/Yoast/javascript/pull/57) is in the `devlop` branch.

This is due to a circular dependency.
The `@yoast/search-metadata-preview` package depends on `yoast-components` and `yoast-components` depends on `@yoast/search-metadata-preview`. This leaves us unable to make this work right now.
The desired outcome would be as follows: `yoast-components` depends on `@yoast/search-metadata-preview` (in order to export it), which depends on `@yoast/components`(for its reuable components).

PR #78, opened 2019-03-19
Add @yoast/search-metadata-previews package

diedexx:
Fixes https://github.com/Yoast/javascript/issues/31

Moved the following to a new package (`@yaost/search-metadata-previews`)
- `yoast-components/composites/SnippetEditor`
- `yoast-components/composites/SnippetPreview`

There still are some references to yoast-components, which will trigger an ECYCLE dependency cycle warning. I think we can ignore these until we've moved all code from `yoast-components/composites/plugin/shared` to another package or find another solution.

## Test instructions
- run `yarn test`. This should succeed. 
- run `yarn lint`. This should succeed.


apps/components/SnippetEditorExample.js
```diff
@@ -1,12 +1,12 @@
 // External dependencies.
 import React, { Component } from "react";
 
+// Internal dependencies.
 import ExamplesContainer from "./ExamplesContainer";
 import { assessments } from "yoastseo";
-const { MetaDescriptionLengthAssessment } = assessments.seo;
+import { SnippetEditor } from "@yoast/search-metadata-previews";
```
bintzandt: This is not really an internal dependency 🤷‍♂️

packages/search-metadata-previews/SnippetEditor/components/ModeSwitcher.js
```diff
@@ -2,14 +2,14 @@
 import React from "react";
 import styled from "styled-components";
 import { __ } from "@wordpress/i18n";
+import PropTypes from "prop-types";
 
 // Internal dependencies.
-import { Button } from "../../Shared/components/Button";
+import { Button } from "yoast-components/composites/Plugin/Shared/components/Button";
+import { SvgIcon } from "@yoast/components";
```
bintzandt: Same comment as before, I don't think that these import are really internal.

packages/search-metadata-previews/SnippetEditor/components/SnippetEditor.js
```diff
@@ -9,7 +9,7 @@ const { MetaDescriptionLengthAssessment, PageTitleWidthAssessment } = assessment
 import { helpers } from "yoastseo";
 const { measureTextWidth } = helpers;
 
-import stripSpaces from "yoastseo/src/stringProcessing/stripSpaces";
+import { string } from "yoastseo";
```
bintzandt: I like this change :)

packages/search-metadata-previews/SnippetPreview/components/HelpTextWrapper.js
```diff
@@ -5,11 +5,11 @@ import styled from "styled-components";
 import uniqueId from "lodash/uniqueId";
 
 // Internal dependencies
-import HelpText from "../../Shared/components/HelpText";
-import { colors, rgba } from "@yoast/components/style-guide";
-import { Button } from "../../Shared/components/Button";
+import HelpText from "yoast-components/composites/Plugin/Shared/components/HelpText";
```
bintzandt: I'm guessing that this reference needs to change after the `Plugin/Shared/components` are merged into `@yoast/components`.

packages/search-metadata-previews/index.js
```diff
@@ -0,0 +1,2 @@
+export * from "./SnippetEditor";
```
bintzandt: I am not sure, but shouldn't we assign names to these exports?

packages/search-metadata-previews/jest/setupTests.js
```diff
@@ -0,0 +1,5 @@
+import "jest-styled-components";
+import Enzyme from "enzyme";
+import EnzymeAdapter from "enzyme-adapter-react-16";
```
bintzandt: Do we still need `Enzyme` here? I found that my tests were also working when I removed enzyme.

packages/yoast-components/index.js
```diff
@@ -66,16 +66,15 @@ export {
...
 export * from "./forms";
 export * from "./composites/Plugin/ContentAnalysis";
+export * from "@yoast/search-metadata-previews";
 export { default as utils } from "./utils";
 export { localize } from "./utils/i18n";
 export { setTranslations } from "./utils/i18n";
 export { translate } from "./utils/i18n";
 export * from "./composites/Plugin/DashboardWidget";
-export { replacementVariablesShape, recommendedReplacementVariablesShape } from "./composites/Plugin/SnippetEditor/constants";
+export { replacementVariablesShape, recommendedReplacementVariablesShape } from "../search-metadata-previews/SnippetEditor/constants";
```
bintzandt:
```suggestion
export { replacementVariablesShape, recommendedReplacementVariablesShape } from "@yoast/search-metadata-previews/SnippetEditor/constants";
```

packages/search-metadata-previews/SnippetPreview/components/HelpTextWrapper.js
diedexx: Yup

packages/search-metadata-previews/jest/setupTests.js
diedexx: haha I found the exact opposite. There are only one or two tests that require it

PR #87, opened 2019-03-20
Move most of the 'shared' components.

### LINKED ISSUE #71: Move the plugin/share/components to the @yoast/components package
## Summary

Move the `yoast-components/composites/Plugin/Shared/components` to the `@yoast/components` package

## Explanation

We have to move the components in `yoast-components/composites/Plugin/Shared/components` to the `components` package. For the most of the components this would't be a problem. However there will be components that already exists with the same name. These should be discussed with an architect.

andizer:
Fixes #71 

Most of the components in yoast-components/composites/Plugin/Shared are moved to the components package.

There are a couple of components remaining. These are based on a button component. See #88


packages/algolia-search/SearchResultDetail.js
```diff
@@ -3,11 +3,10 @@ import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 import { defineMessages, injectIntl, intlShape } from "react-intl";
-import { YoastButton, SvgIcon } from "yoast-components";
+import { YoastButton, SvgIcon } from "@yoast/components";
 import { colors, breakpoints } from "@yoast/components/style-guide";
 
-import { YoastLinkButton } from "yoast-components/composites/Plugin/Shared/components/YoastLinkButton";
-import { makeOutboundLink } from "yoast-components/utils/makeOutboundLink";
+import { YoastLinkButton, makeOutboundLink } from "@yoast/components";
```
bintzandt: I would throw everything in one single import from `@yoast/components`.

packages/components/jest/setupTests.js
```diff
@@ -0,0 +1,6 @@
+import "raf/polyfill";
+import "jest-styled-components";
+import Enzyme from "enzyme";
+import EnzymeAdapter from "enzyme-adapter-react-16";
+
+Enzyme.configure( { adapter: new EnzymeAdapter() } );
```
bintzandt: We might not need Enzyme here. Not sure, so please check for yourself :)

PR #93, opened 2019-03-20
Make screen readers correctly announce Facebook preview site and author name

### LINKED ISSUE #89: Screenreader doesn't handle facebook sitename and author well
The following Facebook sitename and author are pronounced like 'Sitename dot C O M vertical line B Y John D O E' by VoiceOver:
Schermafbeelding 2019-03-20 om 10 19 00

Adding `aria-hidden="true"` to the span that contains the `|` does solve the 'vertical line' problem, but since we also have to deal with the capital letters problem, adding an `aria-label` might be a better option.

Parent: https://github.com/Yoast/javascript/issues/47

afercia:
Fixes #89 

- adds a screen-reader-text with lowercase text for screen readers
- hides the uppercase text from screen readers

Initially, I thought to use an `aria-label` but on a non-interactive element `aria-label` makes also screen reader announce the element has "group" (that is, it assign it an ARIA role).

Finally opted for a screen-reader-text, sees to me the cleanest way to provide a lowercase string to screen readers. Also, improves a bit the conditional rendering of the author name.

Screenshot 2019-03-20 at 13 57 28

(notice above the lowercase text gets announced vs. the visible uppercase text)

### DIFF
--- a/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
+++ b/packages/yoast-components/composites/Plugin/SocialPreviews/Facebook/components/FacebookSiteAndAuthorNames.js
@@ -31,7 +31,7 @@ const FacebookSiteAndAuthorNamesSeparator = styled.span`
  * @returns {React.Element} The rendered element.
  */
 function renderFacebookAuthorName( authorName ) {
-	/* translators: the context is: SITE | By AUTHOR */
+	/* Translators: the context is: SITE | By AUTHOR */
 	const by = __( "By", "yoast-components" );
 
 	return (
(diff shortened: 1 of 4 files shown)


afercia:
Looking into failing test.

IreneStr:
CR 👍

johannadevos:
Acceptance: whether or not the text is read out correctly, depends on how I navigate through the elements.

* When I navigate through the elements by using Control Option Right/Left, the text is read out correctly.
* When I navigate through the elements simply by using the Up arrow or Down arrow, starting at `FacebookPreview Landscape` and going down, first all the individual elements are read out correctly, one by one ("Sitename.com", "by", "John", "Doe"). However, I then land on the element as a whole (`SITENAME.COM | BY JOHN DOE`), and I hear: "Sitename dot C O M vertical line John Doe"

afercia:
Thanks @johannadevos. Not an actual problem as the VoiceOver [standard keyboard shortcuts](https://www.apple.com/voiceover/info/guide/_1131.html) for navigation use the "VO" (Control+Option) modifier. The single Up and Down arrows are not standard in VoiceOver.

All screen readers allow also to read "by word" or "by character" and in this cases aria-hidden may be ignored but there's really nothing we can do about that :)

johannadevos:
@afercia Alright, then acceptance is 👍 and I will merge this PR :)

PR #96, opened 2019-03-20
Improve the layout of the AlgoliaSearch package.

bintzandt:
# Summary
Improve the layout of the AlgoliaSearch package.

**Note: The `babel.config.js` should be removed in the future. However, `jest` was not willing to work without it. Also: moving it to `tools/babel` (our naming convention) does not work as expected.**


andizer:
CR 👍

PR #97, opened 2019-03-21
Improve components structure

andizer:
Moves all files to the new folder structure


bintzandt:
CR: 👍
Acc: 👍 

Can be merged if the build succeeds.

PR #98, opened 2019-03-21
Improve configuration wizard structure

bintzandt:
# Summary
* [x] Improve the structure of the configuration wizard.
* [x] Improve the names of files in the configuration wizard.
* [x] Remove unused files in the configuration wizard.
* [x] Add eslint and try to remove as much errors as possible.

# Technical decisions
I found it hard to find a suitable location for some files. These are config-files for the api. I currently put them under `/tools/config/`. There were also some interfaces that are only used in those config files.  I put them in `/tools/config/interfaces/`. Both choices are up for discussion since there is currently no standard for them.


packages/configuration-wizard/src/index.js
```diff
@@ -0,0 +1,22 @@
+/* Yoast imports */
+import { Explanation } from "@yoast/components";
```
andizer: I think this isn't needed, because we should load it from `@yoast/components.`

packages/configuration-wizard/src/index.js
```diff
@@ -0,0 +1,22 @@
...
+import { decodeHTML, sendRequest } from "@yoast/helpers";
```
andizer: We shouldn't expose this in this package.

andizer:
Awesome 👍

PR #100, opened 2019-03-21
Refactor search metadata previews into agreed upon package structure

diedexx:
# Test instructions
yarn test and yarn lint should still succeed


andizer:
I think we should do a better restructure:
- Only expose components by `src/index.js` and we shouldn't have index.js files in subdirectories.
- I would combine the `constants.js` from `SnippetEditor` and `SnippetPreview in one single file`
- I would put the files `SnippetEditor/serialization.js`, `SnippetEditor/selection.js`, `SnippetEditor/replaceText` and `SnippetEditor/positionSuggestions` in a helper directory that is placed in the root of the package.
- The files that are located in the components directories of `SnippetEditor` and `SnippetPreview` can be moved to the root of both dirs.

By doing this the file structure will be more flatten.

packages/yoast-components/index.js
```diff
@@ -74,7 +74,6 @@ export { localize } from "./utils/i18n";
 export { setTranslations } from "./utils/i18n";
 export { translate } from "./utils/i18n";
 export * from "./composites/Plugin/DashboardWidget";
-export { replacementVariablesShape, recommendedReplacementVariablesShape } from "@yoast/search-metadata-previews/SnippetEditor/constants";
```
andizer: This can break things

diedexx: As discussed in person: It will not break because these modules are already exported by https://github.com/Yoast/javascript/pull/100/files/7d2a5cd0dcf459a557851b129aca7a695948cb3d#diff-9e4399436200d7ba9ddad22eb57ab48aR71

packages/search-metadata-previews/src/index.js
```diff
@@ -0,0 +1,9 @@
+// Snippet preview exports.
+export { default as FixedWidthContainer } from "./snippet-preview/FixedWidthContainer";
+export { default as HelpTextWrapper } from "./snippet-preview/HelpTextWrapper";
+export { default as SnippetPreview } from "./snippet-preview/SnippetPreview";
+
+// Snippet editor exports.
+export { default as ReplaceVarEditor } from "./snippet-editor/ReplacementVariableEditor";
```
atimmer: Can we change `ReplaceVarEditor` to `ReplacementVariableEditor`? We can keep it the old way in `yoast-components`.

diedexx:
>  Also, we might want to move those things to @yoast/components.

This PR contains a lot of change already. I prefer not to add any more, as this change would most likely require us to split up the shared file in a generic yoast/components worthy part and a snippet-editor/preview specific part.

bintzandt:
At the moment, none of the components in `shared.js` are used outside of the `search-metadata-previews`. Therefore I think that it is not worth to move them to the `@yoast/components` package. At least not in this PR. 

As to the second remark of @atimmer: it might be possible to move all of the `.codeclimate.yml` and `.editorconfig` files to the root of the repository. I think it might be best to create a separate issue for that so the scope of this issue remains the same.

Keeping in mind the remarks above, CR: 👍

bintzandt:
ACC: 🚧 

The `components` app does not work. This seems to be due to an unrecognized import in the `UIControlsWrapper.js`.

diedexx:
This is the case on develop as well. Creating an issue later. Merging for now

PR #101, opened 2019-03-21
Allow developers to use their own branch for premium-configuration.

### LINKED ISSUE #16: Add a message to yarn test if the premium-configuration is not installed

bintzandt:
This should fix the problem that @nataliashitova had.

Waiting for her to verify that the problem is fixed, can go through the process of CR though.

Closes #16


atimmer:
CR Done, this PR is ready for acceptance 👍🏻

PR #103, opened 2019-03-21
Use the right keyphrase boundaries for morphology languages

manuelaugustin:
## Summary

* [not-user-facing] Makes sure that all languages with morphology support use the correct keyphrase boundaries.

* Previously, the check for whether to use the specific keyphrase boundaries for languages with morphology support was set hard-coded to English. Since that is not the only language with morphology support, we now make use of the helper function `getLanguagesWithMorphologySupport`.

## Test instructions

* In the example, set a keyphrase and add a text that's long enough to trigger the keyphrase density assessment (>150 words).
* Add `console.log( this._keywordDensity )` on line 117 of `KeywordDensityAssessment`. You'll need this to montor the calculated keyphrase density, as it's not apparent from the feedback strings.
* Set your locale to `de_DE`. Add the keyphrase often enough in the text to get a density between 3 and 3.5. If you're within this range, you should get a green score.
* Now set your locale to one for which we don't have morphology support, e.g. `nl_NL`. The keyphrase density should stay the same, but you should now get a red bullet.

Fixes #


nataliashitova:
CR and acceptance 👍

PR #104, opened 2019-03-21
Remove the Explanation component.

### LINKED ISSUE #67: Change class name for the Explanation component
The `Explanation` component is moved from the OnBoardingWizard directory to the `@yoast/components`. This is because a not onboarding wizard components is using it. The only thing is that the component has a class name `yoast-wizard-input__explanation` and that should be changed.

I think we should pass the class name as a property to the component.

bintzandt:
As discussed in #67 we want to remove the Explanation component because it is very simple and only used in a handful of places.

* [x] Remove component
* [x] Update snapshots
* [x] Update usages

This PR removes the component.

_It might be wise to hold off from merging this until #97 is fixed._

Fixes #67


packages/components/composites/Textfield.js
```diff
@@ -58,7 +57,11 @@ class Textfield extends React.Component {
 						hasFocus={ this.props.hasFocus }
 						value={ this.props.value }
 					/>
-					<Explanation text={ this.props.explanation } />
+					{ this.props.explanation && (
+						<p className="yoast-wizard-input__explanation">
```
andizer: The classname can removed here :) Because this field is not in the configuration wizard context.

packages/components/composites/Textfield.js
```diff
@@ -75,7 +78,11 @@ class Textfield extends React.Component {
 					autoComplete={ this.props.autoComplete }
 					optionalAttributes={ this.optionalAttributes.field }
 				/>
-				<Explanation text={ this.props.explanation } />
+				{ this.props.explanation && (
+					<p className="yoast-wizard-input__explanation">
```
andizer: See above

PR #108, opened 2019-03-22
Use custom researcher in taxonomy assessor

nataliashitova:
## Summary
* [not-user-facing] Fixes the bug when custom `researcher` would not be passed to the taxonomy assessor, which would result in morphological analysis not run on taxonomy pages (concerns English and German)

## Testing instructions
* Use the local testing environment to check if the morphology data is being used for taxonomy analysis.

Fixes https://github.com/Yoast/wordpress-seo-premium/issues/2305

### DIFF
--- a/packages/yoastseo/src/relatedKeywordTaxonomyAssessor.js
+++ b/packages/yoastseo/src/relatedKeywordTaxonomyAssessor.js
@@ -11,10 +11,11 @@ import FunctionWordsInKeyphrase from "./assessments/seo/FunctionWordsInKeyphrase
  * Creates the Assessor used for taxonomy pages.
  *
  * @param {object} i18n The i18n object used for translations.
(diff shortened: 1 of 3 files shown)


manuelaugustin:
CR & acceptance 🚧

As far as I can see it's necessary to implement the same functionality also in the `RelatedKeywordTaxonomyAssessor`.

nataliashitova:
@manuelaugustin, sorry, overlooked it.

manuelaugustin:
CR & acceptance 👍

Yay for fixing this nasty little 🐛!

@nataliashitova Could you update the changelog entry? Because actually it's a user-facing change. Something like `Fixes a bug that impeded recognition of word forms for keyphrases on taxonomy pages.` The current description can best be moved to `Relevant technical choices` I think.

PR #124, opened 2019-03-25
Move course cards to @yoast/components

### LINKED ISSUE #82: Move course-overview file to @yoast/components
## Summary

We want to have structured packages for all our components. To achieve this, we need to move a course-overview file to another package.

## Explanation

This course-overview file needs to be moved to the new @yoast/components package:
javascript/packages/yoast-components/composites/CoursesOverview/Card.js

## Technical decisions

The version of the new package needs to be 0.1.0

diedexx:
Fixes #82


packages/components/src/CardDetails.js
```diff
@@ -0,0 +1,177 @@
...
+
+	/**
+	 * Returns the correct Action Block based on whether an item is a bundle or a single course.
+	 *
+	 * @param {string} buttonType The type of the button. Either regular or sale.
+	 * @param {boolean} isBundle True when the it's a bundle.
+	 *
+	 * @returns {ReactElement} The ActionBlock component.
+	 */
+	getActionBlock( buttonType, isBundle ) {
```
bintzandt: I would move this function outside of the `CardDetails` class (and also move it before the class since the function is now called before it is defined, which is not a problem in JavaScript AFAIK but just feels weird...).

packages/yoast-components/composites/CoursesOverview/Card.js
```diff
@@ -1,156 +1 @@
...
-};
-
-Card.defaultProps = {
-	className: "",
-	id: "",
-	header: null,
-	banner: null,
-	children: null,
-};
+export { Card as default } from "@yoast/components";
```
bintzandt: According to at @atimmer, we don't need to do this. We are going to rewrite the all the relative usages. We don't need to keep exporting the components on the old places.

packages/yoast-components/composites/CoursesOverview/CardBanner.js
```diff
@@ -1,59 +1 @@
...
-	textColor: PropTypes.string,
-	children: PropTypes.any,
-};
-
-Banner.defaultProps = {
-	backgroundColor: colors.$color_pink_dark,
-	textColor: colors.$color_white,
-	children: null,
-};
+export { CardBanner as default } from "@yoast/components";
```
bintzandt: See comment above 👆

packages/yoast-components/composites/CoursesOverview/CardDetails.js
```diff
@@ -1,186 +1,3 @@
...
-			<OutboundLinkButton href={ this.props.ctaButtonData.ctaButtonUrl } rel={ null }>
-				{ this.props.ctaButtonData.ctaButtonCopy }
-			</OutboundLinkButton>
-			<OutboundInfoLink href={ this.props.courseUrl } rel={ null }>
-				{ this.props.readMoreLinkText }
-			</OutboundInfoLink>
-		</ActionBlock>;
-	}
-}
+import { CourseDetails as CardDetails } from "@yoast/components";
```
bintzandt: See comment above 👆

packages/yoast-components/composites/CoursesOverview/Card.js
diedexx: It feels kinda weird to me to remove BC code and alter the consumers which were using incorrect imports, only to have us replace the corrected import by an import of the new package later still.

What I'm trying to say: If I leave these lines, nothing has to change up until the point when we're updating dependencies to the new packages in the consuming applications.
If we remove this, we'll have to do extra work to make sure the consuming applications still work

atimmer: Fair enough, it doesn't hurt to have these files.

atimmer: @bintzandt This is different from what I said yesterday, but I changed my mind by @diedexx's comment.

packages/yoast-components/composites/CoursesOverview/CardBanner.js
andizer: Is this resolved?

packages/yoast-components/composites/CoursesOverview/CardDetails.js
andizer: Same question

andizer:
Can you fix the build?

PR #132, opened 2019-03-26
Add ESLinting to all packages

atimmer:
* Not all package had linting yet. This PR makes sure they are linted.
* I have consolidated all `.eslintrc` into the one in the root.
* I have added test running to the `helpers` package.


bintzandt:
CR:
- Some packages still have an `.eslintignore` file. I don't know if this can be moved to the root as well or if it is necessary that they remain where they are.
- `yoast-social-previews` and `yoast-components` still have their own `.eslintrc`. I don't know if this is by accident or intentional.

If both things above are intentional then the CR is OK 👍

atimmer:
> yoast-social-previews and yoast-components still have their own .eslintrc. I don't know if this is by accident or intentional.

This is intentional, because those packages are being deprecated and I felt I shouldn't spend time on these.

PR #148, opened 2019-03-27
Fix sass in configuration-wizard.

### LINKED ISSUE #141: Fix sass in configuration wizard
## Summary

The path to the sass file for the configuration wizard needs to be adjusted.

## Explanation

In the JavaScript monorepo the sass file for the configuration wizard has been removed and needs to be created in @yoast/configuration-wizard/src
In the plugin repo the path to this file needs to be adjusted so the file can be found.
Check if there are other locations where this file is used.

## Technical decisions

2 PRs are needed for this. One on the JavaScript monorepo and one on the plugin repo.

bintzandt:
# Summary
Move the scss file back to the configuration-wizard. Together with the colors.scss which is used by the configuration-wizard. This PR should fix the problems that occur when trying to build the configuration-wizard.

Fixes #141
Related Yoast/wordpress-seo#12531


packages/configuration-wizard/src/index.js
```diff
@@ -5,6 +5,9 @@ import { default as LoadingIndicator } from "./LoadingIndicator";
 import Config from "../tools/config/production-config";
 import apiConfig from "../tools/config/api-config";
 
+export { default as MailchimpSignup } from "../tools/config/interfaces/MailchimpSignup";
```
diedexx: Please don't export tools. If should be exported, it should be in src. Tooling should not be used in src

bintzandt: Yeah, my bad. I thought I needed to export these because they were used in the plugin but I was wrong. Reverting 🔀

diedexx:
Acceptance 👍  Feel free to merge when the build passess

PR #170, opened 2019-03-29
Prevent abbreviations from stemming

### LINKED ISSUE #139: Prevent abbreviations from stemming
## Summary

Short all-caps abbreviations should not be stemmed using general rules. This unwanted behavior leads to stemming, e.g., `CTA` to `ctum` in English.

## Explanation

In consultation with Team Lingo we've decided that all all-caps words 2, 3, and 4 characters long should be prevented from stemming. The result of relevantWords researches should then include abbreviations in a shape of `WordCombination`s, where `word` is an all-caps abbreviation, `stem` is a lower-case abbreviation, `occurrences` is a number of occurrences.

nataliashitova:
## Summary

* Adds functionality that processes abbreviations (2-4 letter long upper-case words) differently than the rest of the words in internal linking suggestions and insights. Abbreviations do not get stemmed, i.e., they are saved in the database as is. Abbreviations are displayed in the Insights as upper-case words.

* For internal linking abbreviations are being determined based on all word occurrences in the text and in the attributes (title, keyphrase, synonyms, metadescription). For insights abbreviations are determined based on the text alone.

## Test instructions

* Use the current branch and the same-name branch on `premium-configuration`.
* Check if all tests pass.
* Use `example` to test that
  * A text without abbreviations gets processed as expected
  * In a text with abbreviations:
    * All non-abbreviations are processed in the old way (get stemmed for the `stem` field, get lowercased for the `word` field)
    * All abbreviations are processed differently (don't get stemmed, get uppercased in the `word` field)
  * Try editing the text though adding and removing words, make sure the analysis does not crashes and the results in the `Prominent words` fields make sense.
* Build a local version of the premium plugin (if [this PR](https://github.com/Yoast/wordpress-seo-premium/pull/2296) has been merged)
  * Run the dev server to see debug messages and print-out of results of researches
  * Use branch `feature/internal-linking` on `wordpress-seo-premium`.
  * **Not-so-nifty-trick**: Because of the fact that we cannot use custom versions of the `premium-configuration` when building the plugin locally, you will need to disable all functionality that relies on the rules added in the `feature/internal-linking`.
    * In `/yoastseo/src/morphology/english/determineStem.js` comment out lines 96 to 99
    * In `/yoastseo/src/morphology/english/getAdjectiveForms.js` comment out lines 82 to 112
  * Link `yoastseo` and `yoast-components` to your `wordpress-seo-premium` branch and build `javascript`, `wordpress-seo-premium` and the `premium` folder of the latter.
  * Open the local version of the plugin and add a new post.
  * Note down the `id` of the post and check which words get saved in the database. To see the words that get saved, you can use a query 
```
SELECT * FROM `wp_term_relationships`
INNER JOIN `wp_term_taxonomy` ON `wp_term_taxonomy`.`term_taxonomy_id` = `wp_term_relationships`.`term_taxonomy_id`
INNER JOIN `wp_terms` ON `wp_term_taxonomy`.`term_id` = `wp_terms`.`term_id`
WHERE `object_id` = 'XXXXXXXXXXX';
```
*XXXXXXXXXXX should be your post id*
  * Check that
    * A text without abbreviations gets processed as expected
    * In a text with abbreviations:
      * All non-abbreviations are processed in the old way (get stemmed for the `stem` field, get lowercased for the `word` field)
      * All abbreviations are processed differently (don't get stemmed, get uppercased in the `word` field)

## UI changes

## Quality assurance

Fixes #139


packages/yoastseo/src/stringProcessing/relevantWords.js
```diff
@@ -164,52 +194,45 @@ function computeRelevantWords( words, language, morphologyData ) {
...
  *
- * @param {Object} attributes                  The attributes to process
- * @param {string} attributes.keyphrase        The keyphrase of the paper.
- * @param {string} attributes.synonyms         The synonyms of the paper.
- * @param {string} attributes.metadescription  The metadescription of the paper.
- * @param {string} attributes.title            The title of the paper.
- * @param {string[]} attributes.subheadings    The subheadings of the paper.
- * @param {string} language                    The language of the paper.
- * @param {Object} morphologyData              The morphologyData available for the language of the paper.
+ * @param {string}      attributes       The attributes to process.
```
hansjovis: I would make this parameter an array of word strings, instead of a single text string, and adapt the code accordingly.

This would make it easier to re-use this function when we make the step to using a tokenizer / linguistic tree.

packages/yoastseo/src/stringProcessing/relevantWords.js
```diff
@@ -164,52 +194,45 @@ function computeRelevantWords( words, language, morphologyData ) {
...
 } );
 
 
 /**
  * Gets relevant words from the paper text.
  *
- * @param {string} text The text to retrieve the relevant words of.
- * @param {string} language The paper's language.
- * @param {Object} morphologyData The morphologyData available for the language of the paper.
+ * @param {string}      text            The text to retrieve the relevant words from.
```
hansjovis: Same here. I would replace this with an array of word strings.

packages/yoastseo/src/stringProcessing/relevantWords.js
nataliashitova: Agreed

packages/yoastseo/src/stringProcessing/relevantWords.js
nataliashitova: I think that for now, it's an overkill. It's a fairly straightforward change to implement once we have the linguistic tree.

hansjovis:
Acceptance test: _pending_

* All unit tests pass!
* Acceptance test in the local development tool pass!
* Local premium tests are still pending...

Issue #160, opened 2019-03-27
Refused to apply style error in components example


IreneStr:
1. Navigate to `javascript/apps/components`
2. Run `yarn install` and `yarn start`
3. Go to `localhost:3333`
4. Open your console.
5. See the following error:
<img width="593" alt="Schermafbeelding 2019-03-27 om 16 18 17" src="https://user-images.githubusercontent.com/17744553/55088418-fdd90a00-50ab-11e9-9c37-52f01ad02e3c.png">

PR #172, opened 2019-03-29
Use Webpack to load css

diedexx:
Fixes https://github.com/Yoast/javascript/issues/160

## Summary 
We tried to load plugin.css from the node_modules with a static path reference.
Because of lerna+yarn+workspaces and their hoisting, we don't know the exact path in every situation, so we have to use Webpack imports to resolve the location of the file.

As described by https://jestjs.io/docs/en/webpack.html, I now mock all imported .css files. This is because we now load the draft-js plugin style trough webpack instead of through an HTML import.

## Test instructions
* yarn install and yarn start in `javascript/apps/components`
* Go to the snippet preview tab
* Click Edit snippet
* See that the replace variables have purple styling and there is the right amount of padding to it. (see screenshot)

**new:**
image
**old:**
image

## UI changes

## Quality assurance

Fixes #


andizer:
CR done 👍

bintzandt:
ACC: 👍

PR #192, opened 2019-04-02
The example app now compiles its own css

### LINKED ISSUE #165: The apps/components should have its own CSS and SCSS (for now).
## Explanation
While we are busy updating everything to `styled-components` the example-components app should have its own scss and css. At the moment, there is no way to update the app's CSS. These files were simply copied from yoast-components.

For time being, we should move all the css/scss that the app needs to a folder in the app, and add `grunt` options so we can build it. 

## Technical decisions
This is a dirty fix for now. In the future, we won't have sass anymore, so the files can simply be removed.

### LINKED ISSUE #173: Remove Yoast prefix from 5 components in @yoast/components
## Explanation

The prefix is absolutely redundant and clarity reducing.

bintzandt:
## Summary
The example app should be able to update its css.

Also: this PR fixes the redundant `yoast` prefix.

Fixes #165
Fixes #173


packages/components/src/KeywordSuggestions.js
```diff
@@ -74,7 +74,7 @@ const KeywordSuggestions = ( { relevantWords, keywordLimit } ) => {
 	return ( <WordList
 		title={ __( "Prominent words", "yoast-components" ) }
 		words={ keywords }
-		classNamePrefix="yoast-keyword-suggestions"
+		classNamePrefix="keyword-suggestions"
```
diedexx: Have you verified that the class `yoast-keyword-suggestions` is not used in the plugin?

packages/components/src/Loader.js
```diff
@@ -14,7 +14,7 @@ const Loader = ( { className } ) => {
 		className += " ";
 	}
 
-	className += "yoast-loader";
+	className += "loader";
```
diedexx: Have you verified that the class `yoast-loader` is not used in the plugin?

packages/components/src/StyledSection.js
```diff
@@ -109,7 +109,7 @@ StyledSection.propTypes = {
 };
 
 StyledSection.defaultProps = {
-	className: "yoast-section",
+	className: "section",
```
diedexx: Have you verified that the class `yoast-section` is not used in the plugin?

packages/components/src/YoastModal.js
```diff
@@ -51,36 +51,36 @@ class BaseYoastModal extends React.Component {
 				onRequestClose={ this.props.onClose }
 				role="dialog"
 				contentLabel={ this.props.modalAriaLabel }
-				overlayClassName={ `yoast-modal__overlay ${ this.props.className }` }
-				className={ `yoast-modal__content ${ this.props.className }` }
+				overlayClassName={ `modal__overlay ${ this.props.className }` }
```
diedexx: Have you verified that the class is not used in the plugin? Same goes for all other changes in this file

apps/components/webpack.config.js
```diff
@@ -10,6 +10,12 @@ module.exports = {
 		// Polyfill
 		"babel-polyfill",
 
+		// Compile the @yoast/configuration-wizard scss
+		"@yoast/configuration-wizard/src/configuration-wizard.scss",
```
diedexx: Out of curiosity, where does this CSS end up?

bintzandt: I believe that it ends up in a temporary directory. The CSS is not stored anywhere and is directly inserted in the HTML that is used for displaying the app.

packages/components/src/StyledSection.js
bintzandt: This is used in the plugin (just found out). Should I revert this change or change the usage of this class in the plugin?

packages/components/src/KeywordSuggestions.js
bintzandt: 👍

packages/components/src/Loader.js
bintzandt: 👍

packages/components/src/YoastModal.js
bintzandt: Only in outdated `scss` I believe. Outdated as in: we created a Styled Component from the Modal right?

bintzandt:
@diedexx I also changed the classname of `yoast-section` to `section` in the plugin (see Yoast/wordpress-seo#12574). This can be merged after the changes in this PR.

packages/components/src/KeywordSuggestions.js
andizer: I suggest to revert this change. In WordPress context it is better to have prefixed classnames to prevent conflicts with other plugins or themes.

packages/components/src/Loader.js
andizer: Please revert this change. Classname `loader` is too generic in plugin context.

packages/components/src/StyledSection.js
andizer: I would love to see this reverted. Same reasons as mentioned before.

packages/components/src/YoastModal.js
andizer: See earlier comments.

packages/components/src/YoastModal.js
```diff
@@ -51,36 +51,36 @@ class BaseYoastModal extends React.Component {
...
+				className={ `modal__content ${ this.props.className }` }
```
andizer: Same...

packages/components/src/YoastModal.js
```diff
@@ -51,36 +51,36 @@ class BaseYoastModal extends React.Component {
...
 				appElement={ this.props.appElement }
-				bodyOpenClassName="yoast-modal_is-open"
+				bodyOpenClassName="modal_is-open"
 			>
 				{
 					this.props.heading &&
-						<StyledHeading className="yoast-modal__title">{ this.props.heading }</StyledHeading>
+						<StyledHeading className="modal__title">{ this.props.heading }</StyledHeading>
```
andizer: ☝️

packages/components/src/YoastModal.js
```diff
@@ -51,36 +51,36 @@ class BaseYoastModal extends React.Component {
...
 							onClick={ this.props.onClose }
-							className={ `yoast-modal__button-close-icon ${ this.props.closeIconButtonClassName }` }
+							className={ `modal__button-close-icon ${ this.props.closeIconButtonClassName }` }
 							aria-label={ this.props.closeIconButton }
 						>
 							<SvgIcon icon="times" color={ colors.$color_grey_text } />
 						</StyledCloseIconButton>
 				}
-				<div className="yoast-modal__inside">
+				<div className="modal__inside">
```
andizer: ☝️

packages/components/src/YoastModal.js
```diff
@@ -51,36 +51,36 @@ class BaseYoastModal extends React.Component {
...
 					{ this.props.children }
 				</div>
 				{
 					this.props.closeButton &&
-						<div className="yoast-modal__actions">
+						<div className="modal__actions">
```
andizer: ☝️

packages/components/src/YoastModal.js
```diff
@@ -51,36 +51,36 @@ class BaseYoastModal extends React.Component {
 				onRequestClose={ this.props.onClose }
 				role="dialog"
 				contentLabel={ this.props.modalAriaLabel }
-				overlayClassName={ `yoast-modal__overlay ${ this.props.className }` }
-				className={ `yoast-modal__content ${ this.props.className }` }
+				overlayClassName={ `modal__overlay ${ this.props.className }` }
+				className={ `modal__content ${ this.props.className }` }
 				appElement={ this.props.appElement }
-				bodyOpenClassName="yoast-modal_is-open"
+				bodyOpenClassName="modal_is-open"
```
andizer: ☝️

apps/components/webpack.config.js
```diff
@@ -65,15 +71,18 @@ module.exports = {
...
 				use: [
 					{
 						loader: 'style-loader',
 						options: {
 							insertAt: 'top'
 						}
 					},
-					'css-loader'
+					'css-loader',
+					{
```
diedexx: This could be just the string `sass-loader` if you're not passing any other options, like css-loader. If you need to add options, an object is indeed the way to go. 
(Let's not fail CR because of this)

packages/components/src/input/InputField.js
```diff
@@ -4,19 +4,6 @@ import styled from "styled-components";
...
-`;
-
-export const YoastInputLabel = styled.label`
-	font-size: 1em;
-	font-weight: bold;
-	margin-bottom: 0.5em;
-	display: block;
-`;
-
 export const YoastInputField = styled.input`
```
diedexx: Please export `InputField` instead of `YoastInputField`.

packages/components/src/SynonymsInput.js
```diff
@@ -3,7 +3,9 @@ import PropTypes from "prop-types";
 import uniqueId from "lodash/uniqueId";
 import styled from "styled-components";
 
-import { YoastInputContainer, YoastInputField, YoastInputLabel } from "./YoastInput";
+import { YoastInputContainer } from "./input/InputContainer";
```
diedexx: Please remove the prefix from all these exports

packages/components/src/index.js
```diff
@@ -64,13 +64,13 @@ export { default as Toggle } from "./Toggle";
...
 export { YoastButton } from "./buttons/YoastButton";
-export { YoastInputField } from "./YoastInput";
+export { YoastInputField } from "./input/InputField";
 export { YoastLinkButton } from "./buttons/YoastLinkButton";
-export { default as YoastLogo } from "./YoastLogo";
-export { default as YoastModal } from "./YoastModal";
-export { default as YoastSeoIcon } from "./YoastSeoIcon";
-export { default as YoastTabs } from "./YoastTabs";
-export { default as YoastWarning } from "./YoastWarning";
+export { default as YoastLogo } from "./Logo";
```
diedexx: Since this is a new package, let's not export these with the Yoast prefix. Right now, it's relatively easy to rename exports. This will prove more challenging the longer we wait.

packages/components/src/input/InputContainer.js
```diff
@@ -0,0 +1,8 @@
+// External dependencies.
+import styled from "styled-components";
+
+export const YoastInputContainer = styled.div`
```
diedexx: This is prefixed, please remove the prefix

packages/components/tests/YoastModalTest.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { mount } from "enzyme";
 
-import YoastModal from "../src/YoastModal";
+import YoastModal from "../src/Modal";
```
diedexx: prefixed

diedexx:
Acc 👍

PR #197, opened 2019-04-04
Create ParsedPaper class

### LINKED ISSUE #116: Create ParsedPaper class
Related #112 

## Summary
We need a ParsedPaper class that handles data setting and getting.

Spec:

## ParsedPaper class
**Properties**

- **metadata**
_Contextual information related to the text, but not IN the text. E.g. keywords, slugs, meta-description etc_.
- **tree**
_A structured and hierarchical representation of the text_

**Methods**

- **Constructor**( metadata: Object, tree: Node ): ParsedData
- **setTree**( tree: Node )
Sets the tree property.
- **getTree**(): Node
Returns the tree property
- **setMetavalue**( key: string, value: * )
Sets/overwrites a key-value pair to the metadata
- **getMetavalue**( key: string ): *
Gets a specific value for the passed key.
- **setMetadata**( metadata: Object )	
Sets/overwrites the metadata
- **getMetadata**(): Object
Returns the metadata object

maartenleenders:
## Summary
[Non-user-facing] This is the ParsedPaper class, which should house the tree and metadata pertaining to tree (such as keyphrases etc).

*

## Test instructions

* The class isn't implemented anywhere yet, and thus cannot be tested (other than verifying that the spec is correct and its tests don't fail).

## UI changes

## Quality assurance

Fixes #116


packages/yoastseo/src/parsedPaper/ParsedPaper.js
```diff
@@ -0,0 +1,81 @@
...
+ * A parsed version of the paper, containing a tree representation of the text,
+ * and parsed metaData pertaining to the tree's context.
+ *
+ * @class ParsedPaper
+ *
+ * @memberOf module:parsedPaper
+ * */
+export default class ParsedPaper {
+	/**
+	 *
```
hansjovis: You could add a simple description, like
```suggestion
	 * Creates a new ParsedPaper.
```

packages/yoastseo/src/parsedPaper/ParsedPaper.js
```diff
@@ -0,0 +1,81 @@
...
+	 */
+	constructor() {
+		this._metaData = {};
```
hansjovis: You can document a (private) property using JSDoc here, when initially declaring it.
E.g.
```js
        /**
         * Stores the metadata (e.g. the keywords, metadescription, title of the document).
         * 
         * @type{Object}
         * @private
         */
		this._metaData = {};
```

packages/yoastseo/src/parsedPaper/ParsedPaper.js
```diff
@@ -0,0 +1,81 @@
...
+		this._tree = {};
```
hansjovis: See my other comment about documenting private properties 😉.

packages/yoastseo/src/parsedPaper/ParsedPaper.js
```diff
@@ -0,0 +1,81 @@
...
+	 * Returns the tree.
+	 *
+	 * @returns {module:parsedPaper/structure.Node|Object} The tree.
+	 */
+	getTree() {
+		return this._tree;
+	}
+
+	/**
+	 * Sets a key-value pair on the metaData object.
```
hansjovis: `metaData` is an internal private property, which means that external developers may lack the context of what this object is when reading the documentation generated by JSDoc.

I would focus the method descriptions more on what they do on a higher level (if you catch my drift 😅).

For example: 'Sets a metadata value on this parsed paper under the given key.'

This would abstract away from the internal workings of the class.

packages/yoastseo/src/parsedPaper/ParsedPaper.js
```diff
@@ -0,0 +1,81 @@
...
+	 *
+	 * @param {string} key   The key for the metaData key-value pair.
+	 * @param {*}      value The value for the metaData key-value pair.
+	 *
+	 * @returns {void}
+	 */
+	setMetaValue( key, value ) {
```
hansjovis: Perhaps `setMetaField` is a better name.
Or `setMetaProperty`, or even `setProperty`, or `setAttribute`?

manuelaugustin:
Returning this to `Needs change` as some Travis tests are failing

manuelaugustin:
Acceptance 👍

I've also added `[Non-user-facing]` to the summary of your PR. (We use this marker when changes don't have any directly visible effects for users.)

PR #199, opened 2019-04-04
Create correct forms for german verbs ending in -t/-et

### LINKED ISSUE #189: German verbs ending in -t/-et aren't stemmed correctly
## Explanation
German verbs ending in `-t`/`-et` aren't stemmed correctly. This concerns, for example, 3rd person singular nouns such as `arbeitet`, `heißt` or `sammelt`. Since they don't get stemmed correctly, we subsequently also don't create the correct forms. We need to stem these words and create additional forms to address this issue.

## Technical decisions
* As we can't reliably predict when a `-t` or `-et` ending is a verb ending, we'll need to create forms for both a stem with the `-t`/`-et`, and a stem without.

Accompanying PRs:
https://github.com/Yoast/javascript/pull/199
https://github.com/Yoast/YoastSEO.js-premium-configuration/pull/33

manuelaugustin:
## Summary
* Improves recognition of German 3rd person singular verb forms (e.g., "arbeitet").

### Creating additional forms for keyphrase recognition
* Since it's not possible to unequivocally determine whether a given form ending in `-t` or `-et` is a 3rd person singular ending or whether it's a stem that happens to end in these letters, we try to make an educated guess. We do this by not creating additional forms for e.g. words with an ending that marks them as a non-3rd person singular form.
* We add additional forms if it's possible that a stemmed word (e.g., "arbeitet") actually needs further stemming (to e.g., "arbeit") and further forms based on this second stem (e.g., "arbeite", "arbeitest", etc.). In some cases, this results in additional nonsense forms (e.g., `student` gets additional forms such as "studene", "studenst" etc.).
### Stemming for internal linking
* Apart from the issue of keyphrase recognition, there is also a problem with stemming for internal linking. Internal linking requires one stem per word. I don't think there's a good way to deal with the ambiguity of stems ending in `t`/`et` to arrive at a single stem that can be used. That's why for now I have only added functionality that does an extra check for verb exceptions. For regular forms and participles, the problem still needs to be resolved. This means that 3rd person singular forms of verbs will not be correctly stemmed and therefore not added to the overall count in internal linking.

## Test instructions

* Set your WordPress site to German.
* Add a text with at least 150 words.
* Add a keyphrase with a German 3rd person singular verb ending in `-t` or `-et` (e.g., "arbeitet").
* Add different forms of the word in your text. You can use a dictionary to find them, e.g. https://de.wiktionary.org/wiki/Flexion:arbeiten.
* Make sure that the keyphrase is recognized.

## UI changes
~* [ ] This PR changes the UI in the plugin. I have added the 'UI change' label to this PR.~

## Quality assurance

Fixes #189


manuelaugustin:
Thanks for your feedback. I've addressed it doing the following:
- I've improved the issue description. Please let me know if it's sufficiently clear now.
- I've put it in the list of issues that still need to be addressed in the overview issue for German morphology improvement https://github.com/Yoast/YoastSEO.js/issues/2160 (I haven't created an issue yet, since we don't know how to address it yet, but see the next point for that.)
- I've made a Trello card where I've specified the problem, so that we can research it more in-depth https://trello.com/c/5Xbdvjj2/57-stemming-german-3rd-person-singular-verb-forms-ending-in-t-et

PR #211, opened 2019-04-08
Wordpress seo premium/2323 related keyphrase not being loaded

hansjovis:
## Summary

* Fixes the issue that the related keyphrase SEO analysis was not being called when the related keyphrases changed. 

* It now specifically checks whether the paper or the related keyphrases have changed since the last analysis.
  * If either has changed, the related keyphrase analysis is executed.
  * Initially, only the paper was checked for changes. However, since the related keyphrases are not stored on the paper, whenever the related keyphrases changed, the analysis was not called.

## Test instructions

**Reproducing the issue**
* Checkout the `release-yoast-seo/11.0` branch
* Start the premium plugin (on the `release/11.0` branch) in development mode.
* Go to a post with no related keyphrase.
* Add a related keyphrase.
* See that the related keyphrase analysis is not run. E.g you get a grey bullet for the related keyphrase analysis and no list of results.

**Testing the fix**
* Checkout this branch `wordpress-seo-premium/2323-related-keyphrase-not-being-loaded`.
* Start the premium plugin (on the `release/11.0` branch) in development mode.
* Go to a post with no related keyphrase.
* Add a related keyphrase.
* Check that the related keyphrase analysis is run. E.g. you get a colored bullet and a list of results for the related keyphrase.
* Change the keyphrase. 
   * Make sure that there would be different results. E.g. change it to a word that is in the introduction of the post, when the original related keyphrase was not.
* Check that the related keyphrase results change.

## UI changes

## Quality assurance

Fixes #https://github.com/Yoast/wordpress-seo-premium/issues/2323

### DIFF
--- a/packages/yoastseo/src/worker/AnalysisWebWorker.js
+++ b/packages/yoastseo/src/worker/AnalysisWebWorker.js
@@ -1,7 +1,7 @@
 // External dependencies.
 import { autop } from "@wordpress/autop";
 import Jed from "jed";
-import { forEach, has, includes, isNull, isObject, isString, isUndefined, merge, pickBy, isEqual } from "lodash-es";
+import { forEach, has, includes, isNull, isObject, isString, isUndefined, merge, pickBy, isEmpty } from "lodash-es";
 import { getLogger } from "loglevel";
 
 // YoastSEO.js dependencies.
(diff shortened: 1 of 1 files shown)


manuelaugustin:
Acceptance test 🚧

As far as I could see, the fix works in part: when I set a related keyphrase or I change the related keyphrase, the results get updated.

What doesn't work, however, is that the results don't get updated if I change the text.

To test, these are the steps you can follow:
* Add the following first paragraph: `This is the first paragraph. It contains a cat.`
* Set the related keyphrase `cat`.
* ✅ The results get updated, I get a green bullet for `Keyphrase in introduction: Well done!`
* Remove `cat` from the text.
* ❌ You still see a green bullet for `Keyphrase in introduction: Well done!`
* ✅ If I change the related keyphrase (e.g., adding a space after `cat` in the related keyphrase field), the analysis updates as expected.

hansjovis:
> What doesn't work, however, is that the results don't get updated if I change the text.

Since we call `analysis` two times, once for the main keyphrase and once for the relate keyphrases, the paper does not have any changes the second time we call the function.

I changed it to only analyze the related keyphrases if they have been given as an argument to the function instead. I opted not to add a "paper has changes" functionality specifically for the related keyphrases, since it would add another lay of complexity to an already complex function and class.

manuelaugustin:
CR & acceptance 👍

Everything works as expected now 🎉

PR #215, opened 2019-04-09
Add irregular adjectives to the German morphology exception check

### LINKED ISSUE #213: Add irregular German adjectives to morphology exception check
We should include very irregular adjectives like _gut_ - _besser_ - _beste_ in the irregular adjectives exception check for German morphology.

agnieszkaszuba:
## Summary
Adds exception checks for very irregular German adjectives.

## Test instructions
* Make sure that correct forms are produced for the following adjectives: gut, viel, hoch 

Fixes #213


packages/yoastseo/src/morphology/german/generateVerbExceptionForms.js
```diff
@@ -56,13 +56,6 @@ const generateFormsPerParadigm = function( morphologyDataVerbs, paradigm, stemme
 	// Some stem types have two forms, which means that a stem type can also contain an array. These get flattened here.
 	stemsFlattened = flatten( stemsFlattened );
 
-	/*
-	 * Sort in order to make sure that if the stem to check is e.g. "gehalt", "halt" isn't matched before "gehalt".
-	 * (Both are part of the same paradigm). Otherwise, if "halt" is matched, the "ge" will be interpreted as preceding
-	 * lexical material and added to all forms.
-	 */
-	stemsFlattened = stemsFlattened.sort( ( a, b ) => b.length - a.length );
```
nataliashitova: This sorting is superfluous because we only use `stemsFlattened` with `includes` in this function. Therefore, @agnieszkaszuba has removed this code.

nataliashitova:
@agnieszkaszuba has added a number of functions to `generateAdjectiveExceptionForms.js` to process exception adjectives. This approach is in line with the previously introduced functions in this file. However, I wonder if it cannot be optimized so that we do not have 10 similar functions that process (slightly differently) different portions of the exception adjectives list. I've made a technical-debt issue to investigate it further: https://github.com/Yoast/javascript/issues/236

manuelaugustin:
CR & acceptance test (i.e., checking that all the tests pass) 👍

I've added some missing function summaries.

PR #241, opened 2019-04-23
Refactors the feature toggle package to make use of a global variable.

hansjovis:
## Summary

* [not-user-facing] Sets the feature flags on the window / worker instead of a 'local' global variable.

* Setting it on the window / worker (on `self`) makes it widely available on the javascript side of the plugin (e.g. also in other packages like `components`).

## Test instructions

* Add the following line to your `wp-config.php` to enable two (non-existing) features:
  `define( 'YOAST_SEO_ENABLED_FEATURES', 'some-feature, another-feature' );`
* Open a file in `components` that gets loaded on a post or page.
  * For example `KeywordSuggestions.js` or `Collapsible.js`.
* Add the following line in the render method or other method that is called inside of the metabox:
  * `console.log( "enabled features", enabledFeatures() );`.
* Open a post or page.
* Open the browser console.
* The console should show: `enabled features: ["some-feature", "another-feature"]`.

## Impact check

  * 

## UI changes

## Quality assurance

Fixes #


packages/feature-toggle/src/index.js
```diff
@@ -1,52 +1,43 @@
...
  *
  * @param {string[]} featureNames   A list of names of the features to enable.
- * @param {string} [namespace=""]   An optional namespace to prepend to each feature name. To avoid possible conflicts between packages.
  *
  * @returns {void}
  */
-const enableFeatures = function( featureNames, namespace = "" ) {
+const enableFeatures = function( featureNames ) {
+	if ( ! self.wpseoFeatureFlags ) {
+		self.wpseoFeatureFlags = featureNames;
```
nataliashitova:
```suggestion
		self.wpseoFeatureFlags = featureNames;
		return;
```

nataliashitova: Because you wouldn't need to loop over the newly created object to see whether featureNames have been defined.

nataliashitova:
Top 👍

PR #249, opened 2019-04-25
Add changelog entries 

afercia:
Adds changelog entries for #8 and #9.

Note: there are some trailing spaces removed, as my editor removes them automatically.


manuelaugustin:
CR 👍
I moved the new entries to the right versions. (We didn't have headings yet for the newest versions).

PR #261, opened 2019-05-01
Make SvgIcon able to accept alternative icon sets

### LINKED ISSUE #229: Make the SvgIcon component able to take an alternate icon set
Context: https://github.com/Yoast/my-yoast/issues/1576#issuecomment-483613494

For better re-use in other applications, the `SvgIcon` component should be able to accept an alternate/additional set of icons. 

At the moment, it uses only the icons defined in its own `icons` object. Exposing this object as an `iconSet` prop would allow to manipulate the default icon set in different ways for example:
- completely replace the default icon set with another one
- add new icons to the default ones

afercia:
## Summary

Makes `SvgIcon` able to accept alternative icon sets by passing the icons object as a prop. This allows to re-use the component in other context where there may be the need to use different icons. See for example https://github.com/Yoast/my-yoast/issues/1576#issuecomment-483613494

- exposing the icon set as an `iconSet` prop allows to manipulate the icons object "in place" where the `SvgIcon` is used
  - add icons to the set
  - remove icons from the set
  - pass a completely different icon set
- note: the `SvgIcon` prop `iconSet` will now contain the whole icons object, not sure this is a problem in terms of performance

## Test instructions
- yarn start the apps components and go to localhost:3333 > SVG Icons tab
- see the examples in the page
- see usage examples in `apps/components/SvgIconsWrapper.js`
- check the icons render correctly where used in the other components
- check the icons still render correctly in the Yoast SEO plugin

## Impact check
  * SVG icons

## UI changes

Fixes #229


afercia:
There's a nice, pending, improvement on MyYoast waiting for this PR, see https://github.com/Yoast/my-yoast/issues/1576#issuecomment-483613494

@IreneStr I'd like to suggest to give this some priority in the process, when there's some room 🙂

Xyfi:
CR 👍

Xyfi:
I'm a bit opposed to the `iconSet` prop and would like to suggest a more permanent way of adding new icons. The current implementation allows you to define `iconSets` in render functions like this:

```
export const MySpecificIcon = () => {
	return (
		<SvgIcon
			icon="random"
			iconSet={ { random: { /* icon data */ } }
		/>
	);
}
```

See how the icon name is directly used to correspond to an icon in the iconSet. Although, I'm not opposed to the idea of the `SvgIcon` component being used for a specific icon. In the following example the icon data is directly passed to the SvgIcon component where a check is performed whether it's a string or an object.

```
export const MySpecificIcon = () => {
	return (
		<SvgIcon
			icon={ /* icon data */ }
		/>
	);
}
```

My preferred solution would be a static function to add new icons to the `SvgIcon` component.

```
SvgIcon.addIcons( "namespace", {
	random: { /* icon data */ },
} );

SvgIcon.addIcons( "namespace", {
	random: { /* icon data */ },
} );
// console.warn => Warning: Icon "random" for namespace "namespace" was already added!
// The above warning should be thrown to enforce a single place where the `SvgIcon` is extended with additional icons.
```

```
export const MySpecificIcon = () => {
	return (
		<SvgIcon
			icon="namespace/random"
		/>
	);
}
```

I'm probably overthinking this.

Xyfi:
Acceptance 👍

PR #306, opened 2019-06-03
Implement Dutch stemming function

### LINKED ISSUE #298: Stem Dutch words
## PR
https://github.com/Yoast/javascript/pull/306

## Summary

We should develop a function that stems Dutch words.

This function will be based on the following algorithm: http://snowball.tartarus.org/algorithms/dutch/stemmer.html

agnieszkaszuba:
## Summary

Stems Dutch words.

Word classes that are stemmed:
- noun plurals (e.g. _katten_) and diminutives (e.g. _katje_)
- all forms of adjectives (e.g. _zachte_, _zachts_, _zachter_, _zachters_, _zachtere_, _zachtst_, _zachtste_ )
- verb infinitives (e.g. _sneeuwen_) and present participles (e.g. _sneeuwend_)

Words classes that are not stemmed correctly
-  present singular (_t)_ and past simple (_-te(n), -de(n_)) verb suffixes (e.g. _sneeuwt_, _sneeuwde_, _sneeuwden_)
- past participles (e.g. _gesneeuwd_)
- word with separable prefixes 
A word like _afwerken_ should in theory get two stems: _werk_ and _afwerk_, since a form like _werkt af_ is a valid form of _afwerken_ . However, only _afwerk_ is currently returned.
- diminutive suffixes -tje (e.g. _biertje_) and -etje (e.g. _sterretje_) not always stemmed correctly (only when we can be sure that the _t/et_ is part of the suffix, not the stem). 
- Words which do not have a clear suffix are not stemmed. This includes forms of irregular words like _is_, _zijn_, _bent_... or words that follow a irregular plural formation pattern, such as words of Latin origin like _musea_. 

Words that are stemmed, but shouldn't (false positives):
- Sometimes, the characters matched by a suffix regex are part of the _stem_, e.g. the -_en_ in _tor*en*_ or the -_end_ in _week*end*_. These words will be incorrectly stemmed. (This problem will be solved in the future, either through having an exception list of words that should not get stemmed, or by adding additional suffixes to make sure each word has its stem ending attached back).

Decisions concerning cases where more than on stem is possible
- When there is a stem with both double and single vowel (e.g. maken -> mak, maak), we return the one with the _double_ vowel
- When there is a stem with both double and single consonant (e.g. paken -> pak, pakk), we return the one with the _single_ consonant
- When there is a stem with both voiced and voiceless consonant (e.g. lezen -> lez, lees), we return the one with the _voiceless_ consonant (so _s_ in this example).
- When the stem ending can be either -ieel or -iël (e.g. officieel/officiële), the stem ending in -ieel is returned.
- The -heden plural suffix is stemmed to -heid.

## Test instructions

* Add more words to the specs and make sure they are stemmed correctly

Fixes #298


manuelaugustin:
In order to make the acceptance test criteria for this PR clear, could you specify the following:
- What classes of words do we currently expect to be stemmed correctly?
- What classes of words do we expect not to be stemmed correctly?
- For words that are stemmed correctly, how do we expect them to be stemmed? (E.g., to what kind of forms are verbs stemmed, how are nouns stemmed that end in a double consonant)

agnieszkaszuba:
> In order to make the acceptance test criteria for this PR clear, could you specify the following:
> 
> * What classes of words do we currently expect to be stemmed correctly?
> * What classes of words do we expect not to be stemmed correctly?
> * For words that are stemmed correctly, how do we expect them to be stemmed? (E.g., to what kind of forms are verbs stemmed, how are nouns stemmed that end in a double consonant)

I added it to the PR summary :)

nataliashitova:
@agnieszkaszuba To be 100% clear in the PR description, please change your subheading `When a word has more than one stem, which one do we return?` to something more confident-sounding, e.g., `Decisions concerning cases where more than on stem is possible`. Otherwise, it might read as if there might be options to discuss further.

packages/yoastseo/spec/morphology/dutch/stemSpec.js
```diff
@@ -0,0 +1,114 @@
+import stem from "../../../src/morphology/dutch/stem";
+
+const wordsToStem = [
```
nataliashitova: Nice list!
Based on the code, the word `collegatjes` will not stem to `collega`, while the words `toren` and `weekend` will be (excessively) stemmed to `toor` and `week`. If so, please adapt the stemmer to accommodate such cases.

I would also like to encourage you to add a comment in the PR description that concerns processing irregular verbs (e.g., `zijn`) and verbs with separable prefixes.

packages/yoastseo/spec/morphology/dutch/stemSpec.js
```diff
@@ -0,0 +1,114 @@
...
+	// -heden gets replaced with -heid (suffix type a1).
+	[ "mogelijkheden", "mogelijkheid" ],
+	// -En suffix preceded by a valid -en ending.
+	[ "vrouwen", "vrouw" ],
+	// A word ending in -en not preceded by a valid -en ending.
+	[ "groen", "groen" ],
```
nataliashitova: Awesome that you decided to include words that should not get excessively stemmed! I would move them to a separate list for readability.

packages/yoastseo/src/morphology/dutch/stem.js
```diff
@@ -0,0 +1,598 @@
...
+	word = undoubleConsonant( word );
+
+	// If stem ends in v, replace it with f. If stem ends in z, replace it with s.
+	word = word.replace( /v$/g, "f" );
+	word = word.replace( /z$/g, "s" );
+
+	// If stem ends with -iël, replace it wth -ieel.
+	word = word.replace( /iël$/g, "ieel" );
+
+	// Turn I and Y back into lower case.
```
nataliashitova: I find this `back` questionable: Actually, all `I`s and `Y` will become small versions of themselves. So, if you receive an abbreviation like `AI` as input, the output of the stemmer will be `Ai`. I think we can solve it by lowercasing all input words either before the stemmer or inside it. In the first case, I would encourage a comment about it in the description of the `stem` function.

packages/yoastseo/spec/morphology/dutch/stemSpec.js
agnieszkaszuba: I added explanations of those cases to the PR descriptions :) And I think _collegatje_ is an example of a word where we cannot know whether the _t_ is part of the stem or not (so whether _-tje_ or _-je_ should be removed). Right now, _-je_ is always removed for those ambiguous cases, but we may be able to improve the stemming using an exception list in the future).

packages/yoastseo/src/morphology/dutch/stem.js
```diff
@@ -0,0 +1,193 @@
...
+ * @param {string} word The stem that needs to be modified.
+ * @param {string} modificationGroup The type of modification that needs to be done.
+ * @returns {string} The modified stem, or the same stem if no modification was made.
+ */
+const modifyStem = function( word, modificationGroup ) {
+	const neededReplacement = modificationGroup.find( replacement => word.search( new RegExp( replacement[ 0 ] ) ) !== -1 );
+	if ( typeof neededReplacement !== "undefined" ) {
+		word = word.replace( new RegExp( neededReplacement[ 0 ] ), neededReplacement[ 1 ] );
+		return word;
+	} return word;
```
hansjovis: Minor thingy:
The first `return word` statement, within the `if` can be removed. 
The modified word will then be returned by the second return statement instead.

```js
if ( typeof neededReplacement !== "undefined" ) {
	word = word.replace( new RegExp( neededReplacement[ 0 ] ), neededReplacement[ 1 ] );
} 
return word;
```

packages/yoastseo/src/morphology/dutch/stem.js
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @param {string} word	 The word from which to delete the suffix.
+ * @param {string} suffixStep 	One of the three steps of deleting a suffix.
+ * @param {number} suffixIndex	 The index of the found suffix.
+ * @param {array} stemModification 	The type of stem modification that needs to be done.
+ * @param {object} morphologyDataNL	 The Dutch morphology data file.
+ * @returns {word} The stemmed and modified word.
+ */
+const deleteSuffixAndModifyStem = function( word, suffixStep, suffixIndex, stemModification, morphologyDataNL ) {
+	if ( String( stemModification ) === "hedenToHeid" ) {
```
hansjovis: Can a particular suffix lead to multiple stem modifications?
I noticed that the `stemModification` is an array, but it is treated as a string in the function:
```js
if ( String( stemModification ) === "hedenToHeid" ) {
```
This works when the modification contains one string: 
```js
console.log( String( [ "hedenToHeid" ] ) ); // Logs: "hedenToHeid"
```
But not when it contains multiple:
```js
console.log( String( [ "hedenToHeid", "vowelDoubling" ] ) ); // Logs: "hedenToHeid,vowelDoubling"
```

I would:
 - Change this function to treat `stemModification` as a string argument.
 - Depending on whether one or multiple stem modifications can be applied:
   - When only one can applied: 
      - Change the stem modification in the morphologyData file as a string, rather than an array.
   - When multiple can applied:
      - Make a separate function that loops over the modifications (in order) and let it call this method on the (changed) word.

packages/yoastseo/src/morphology/dutch/stem.js
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * Deletes the suffix and modifies the stem according to the required modifications.
+ *
+ * @param {string} word	 The word from which to delete the suffix.
+ * @param {string} suffixStep 	One of the three steps of deleting a suffix.
+ * @param {number} suffixIndex	 The index of the found suffix.
+ * @param {array} stemModification 	The type of stem modification that needs to be done.
+ * @param {object} morphologyDataNL	 The Dutch morphology data file.
+ * @returns {word} The stemmed and modified word.
```
hansjovis:
```suggestion
 * @returns {string} The stemmed and modified word.
```

packages/yoastseo/src/morphology/dutch/stem.js
```diff
@@ -0,0 +1,193 @@
...
+ */
+const deleteSuffixAndModifyStem = function( word, suffixStep, suffixIndex, stemModification, morphologyDataNL ) {
+	if ( String( stemModification ) === "hedenToHeid" ) {
+		word = modifyStem( word, morphologyDataNL.stemming.stemModifications.hedenToHeid );
+		return word;
+	} word = word.substring( 0, suffixIndex );
```
hansjovis: I would add this to a new line, perhaps with some blank line padding. Makes the code more readable 😉.
It is easy to overlook that `word` is changed otherwise.

packages/yoastseo/src/morphology/dutch/stem.js
```diff
@@ -0,0 +1,193 @@
...
+
+			if ( matchedRegex ) {
+				const matched = new RegExp( matchedRegex ).exec( word );
+				const suffix = matched[ matched.length - 1 ];
+				const suffixIndex = word.lastIndexOf( suffix );
+
+				if ( suffixIndex >= r1Index ) {
+					return {
+						suffixIndex: suffixIndex,
+						stemModification: ( suffixStep[ suffixClass ].stemModification ),
```
hansjovis:
```suggestion
						stemModification: suffixStep[ suffixClass ].stemModification,
```
The brackets can be removed.

packages/yoastseo/spec/morphology/dutch/stemSpec.js
agnieszkaszuba: When a word ends with letters that are the same as one of the suffixes, but those letters actually belong to the stem, we will run an exception check before stemming to make sure those stem endings are not removed.

The stemmer checks that rely on exception lists will be implemented in the future. See for example the following Trello cards: [stem ending in -en](https://trello.com/c/5UxZQVuo/99-make-a-list-of-nouns-with-stem-ending-in-en), [stem ending in -st](https://trello.com/c/oJIme7QM/102-make-a-list-of-nouns-and-adjectives-with-stem-ending-in-st-ste), also [incorrect vowel doubling](https://trello.com/c/Bf6os3IV/110-fix-incorrect-vowel-doubling-in-the-stemmer)

PR #316, opened 2019-06-18
Create word occurrences bar component

maartenleenders:
## Summary

  * Package(s) involved: components

  * Should the change be included in the package changelog?
    * [ ] No
    * [ ] Yes
  * Should the change be included in one or more plugin changelogs?
    * [ ] No
    * [ ] Free
    * [ ] Premium
    * [ ] Other (please specify)
  * Package changelog item (if applicable): _[behind feature flag]_ Added the WordOccurrences component, which is a new visualisation of the WordList capable of handling improved analysis results.
  * Plugin changelog item (if applicable): _[behind feature flag]_ Changed the visualisation of the Prominent Words list to be more intuitive at a glance.

* Using the WordCloud turned out to not be feasible due to technical shortcomings (Slack communication, product channel June 4th 2019). Therefore we changed the exterior to look like a horizontal bar-plot (see https://github.com/Yoast/wordpress-seo-premium/issues/2297#issuecomment-498749242 ).

## Test instructions

1. Checkout branch in javascript, start the `app/components` app, and verify that the example component works as expected.
2. Run `yarn link-monorepo` in wordpress-seo-premium (trunk), and have the plugin testing tool installed and active.
3. Have a post with a decent amount of text in it, preferably an actual text (not the same word copied 100000 times).

4. If you don't have the `improved internal linking` feature activated in the testing tool, make sure the Insights tab is functioning as expected (as in, nothing changed). Then, activate the feature flag.

5. Verify that the most occurring word has a bar behind it with the full width, and that all fewer occurrences are proportional to that bare (i.e. an occurrence of 30 should be approximately half the length of an occurrence of 60, etc.).

6. Check that the VoiceOver makes sense for the insights tab (cmd + f5 in safari)

7. If you haven't done so, verify that without the `improved internal linking` feature flag active, everything remains the same.

## Impact check

  * Insights tab, prominent words list is now revamped.

## UI changes

## Quality assurance

Fixes https://github.com/Yoast/wordpress-seo-premium/issues/2297


packages/components/src/WordOccurrences.js
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * The WordOccurrences list, that contains words, their occurrence, and bars reflecting their relative occurrence.
+ *
+ * @param {Array} words The array of objects containing words and occurrences.
+ *
+ * @returns {ReactElement} The list of words, their occurrences, and bars.
+ */
+class WordOccurrences extends React.Component {
+	/* eslint-disable-next-line require-jsdoc */
+	constructor( props ) {
```
hansjovis: I would add a JSDoc here.
In this way you can document `props` like in `Collapsible.js`:

https://github.com/Yoast/javascript/blob/b6528b1b4774d29aab5ef08584656d4ce1b95c5f/packages/components/src/Collapsible.js#L202-L206

packages/components/src/WordOccurrences.js
```diff
@@ -0,0 +1,149 @@
...
+	word: PropTypes.string.isRequired,
+	occurrence: PropTypes.number.isRequired,
+	width: PropTypes.string.isRequired,
+};
+
+/**
+ * The WordOccurrences list, that contains words, their occurrence, and bars reflecting their relative occurrence.
+ *
+ * @param {Array} words The array of objects containing words and occurrences.
+ *
```
hansjovis: The `showBeforeList` and `showAfterList` properties should be document as well 😉.

packages/components/src/WordOccurrences.js
```diff
@@ -0,0 +1,149 @@
...
+				{ this.props.showAfterList }
+			</React.Fragment>
+		);
+	}
+}
+
+WordOccurrences.propTypes = {
+	words: PropTypes.array.isRequired,
+	showBeforeList: PropTypes.element,
+	showAfterList: PropTypes.element,
```
hansjovis: Perhaps `header` and `footer` are more suitable names.

packages/components/src/WordOccurrences.js
hansjovis: Also: words can be document more specifically:
```js
/**
 * @param {ProminentWord[]} words The array of prominent words to show.
 */
```

packages/components/src/WordOccurrences.js
```diff
@@ -0,0 +1,149 @@
...
+				min: Math.min( ...allOccurrences ),
+				max: Math.max( ...allOccurrences ),
+			},
+			words: words,
+		};
+	}
+
+	/**
+	 * Renders the WordOccurrences component to the DOM.
+	 * @returns {*} The rendered WordOccurrences component.
```
hansjovis:
```suggestion
	 * @returns {React.Fragment} The rendered WordOccurrences component.
```

packages/components/tests/WordOccurrencesTest.js
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+import renderer from "react-test-renderer";
+
+import WordOccurrences from "../src/WordOccurrences";
+
+const showBeforeList = <p>{ "I'm a before list paragraph" }</p>;
+const showAfterList = <p>{ "I'm an after list paragraph" }</p>;
+const words = [ { _word: "word1", _occurrences: 13 }, { _word: "word2", _occurrences: 8 }, { _word: "word3", _occurrences: 6 } ];
```
hansjovis: The component gets a list of `ProminentWord`s in the plugin. 
It would be better to test if it works with 'real' `ProminentWord`s 😛.

packages/components/src/WordOccurrences.js
```diff
@@ -0,0 +1,149 @@
...
+		};
+	}
+
+	/* eslint-disable-next-line require-jsdoc */
+	static getDerivedStateFromProps( props ) {
+		const words = [ ...props.words ];
+		words.sort( ( a, b ) => {
+			return b._occurrences - a._occurrences;
+		} );
+		const allOccurrences = words.map( wordObject => wordObject._occurrences );
```
hansjovis: Sorry, I just noticed this...
`_occurrences` is a private variable. It will be better to use the public `getOccurences` interface on the `ProminentWord` instance instead, since the inner workings of the class can change in the future.

packages/components/src/WordOccurrences.js
maartenleenders: I agree it might be nice to describe what the params actually are. There is just a lot of overlap between the `@param {type}` setting, and the `React.propTypes` object definition in the bottom.

Did both for now.

packages/components/src/WordOccurrences.js
maartenleenders: Ah yeah, I have not worked with those deeper specifications yet. For me, they don't always add clarity (ProminentWord doesn't yell "object" to me), but down the line specificity is probably better.

packages/components/src/WordOccurrences.js
maartenleenders: Agreed. Copied these from WordList, which is WordOccurrence's legacy mirror component.

packages/components/src/WordOccurrences.js
hansjovis: It also serves as a type hint to PHPStorm and JSDoc. 

This helps PHPStorm with indicating what methods and properties are defined on the given parameters. 

JSDoc in turn can link to the respective documentation of the class (when the documentation is generated).

packages/components/src/WordOccurrences.js
hansjovis: I think this would help when we want to generate documentation in JSDoc in the future. I don't know if JSDoc considers `React.propTypes`...

hansjovis:
Acceptance test: ✅

PR #317, opened 2019-06-18
2432 bump minimum number of word occurrences for prominent word for internal linking.

hansjovis:
## Summary
Bumps the minimum number of word occurrences from 2 to 4 (when extracting prominent words for internal linking).

  * Package(s) involved:

  * Should the change be included in the package changelog?
    * [ ] No
    * [x] Yes
  * Should the change be included in one or more plugin changelogs?
    * [x] No
    * [ ] Free
    * [ ] Premium
    * [ ] Other (please specify)
  * Package changelog item (if applicable): Behind internal linking feature flag (currently disabled by default): Bumps the minimum number of required word occurrences from 2 to 4 (when extracting prominent words for internal linking).
  * Plugin changelog item (if applicable): 

* Updated the test for `getProminentWordsForInternalLinking` to match the changes.

## Test instructions

#### Reproduce the bug.

* Checkout the `feature/internal-linking` branch in this repository.
* Create a new (and thus empty) post.
* Wait 15 seconds.
* You should see suggestions inside the internal linking metabox / sidebar.
* Open the developer console.
* You should see an incoming `runResearch`-request to the `AnalysisWebWorker`.
   * Note its ID-number.
* You should see an outgoing `runResearch:done`-request from the `AnalysisWebWorker` with the same ID-number as you noted down in the previous step.
   * Check that it contains a list of `ProminentWord`s.

#### Check if the bug has been fixed.

* Checkout this branch (`2432-avoid-premature-internal-linking-suggestions`).
* Create a new (and thus empty) post.
* Wait 15 seconds.
* You should **not** see suggestions inside the internal linking metabox / sidebar.
* Open the developer console.
* You should see an incoming `runResearch`-request to the `AnalysisWebWorker`.
   * Note its ID-number.
* You should see an outgoing `runResearch:done`-request from the `AnalysisWebWorker` with the same ID-number as you noted down in the previous step.
   * Check that it contains an empty list.

### Check if all prominent words with less than 4 occurrences are filtered out.

* Checkout this branch (`2432-avoid-premature-internal-linking-suggestions`).
* Open a post.
* Wait 15 seconds.
* You should see an incoming `runResearch`-request to the `AnalysisWebWorker`.
   * Note its ID-number.
* You should see an outgoing `runResearch:done`-request from the `AnalysisWebWorker` with the same ID-number as you noted down in the previous step.
   * Check that it contains a list of `ProminentWord`s, and that all of them have 4 or more occurrences. 

## Impact check

  * The impact should be small, since this only changes the prominent words for internal linking generation. 

## UI changes

## Quality assurance

Fixes https://github.com/Yoast/wordpress-seo-premium/issues/2432

### DIFF
--- a/packages/yoastseo/spec/researches/getProminentWordsForInternalLinkingSpec.js
+++ b/packages/yoastseo/spec/researches/getProminentWordsForInternalLinkingSpec.js
@@ -9,13 +9,13 @@ const morphologyData = getMorphologyData( "en" );
 
 describe( "relevantWords research", function() {
 	it( "does not break if no morphology support is added for the language", function() {
-		const paper = new Paper( "texte et texte", { locale: "fr_FR" } );
+		const paper = new Paper( "texte et texte et texte et texte", { locale: "fr_FR" } );
 
 		const researcher = new Researcher( paper );
 		researcher.addResearchData( "morphology", morphologyData );
 
 		const expected = [
-			new ProminentWord( "texte", "texte", 2 ),
+			new ProminentWord( "texte", "texte", 4 ),
 		];
 
 		const words = prominentWordsResearch( paper, researcher );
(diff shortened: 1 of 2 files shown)


manuelaugustin:
CR & acceptance 👍

Some notes:
- A quick note about the changelog item: I think it makes sense to include this change in the changelog for the library, but point out that it's a change behind the feature flag (which is disabled by default). We did the same when we first added the improved internal linking functionality for part 1. So I added this as the changelog item: `Behind internal linking feature flag (currently disabled by default): Bumps the minimum number of required word occurrences from 2 to 4 (when extracting prominent words for internal linking).`
- I've added an in-line comment to motivate the `4` (`Minimum required occurrences set to 4 in order to avoid premature suggestions of words fro  the paper attributes. These get a times-3 boost and would therefore be prominent with just 1 occurrence.`)
- I think we still need to think a bit more about minima. I think that the fix that we're implementing in this PR is valid and necessary, but perhaps not completely sufficient to avoid premature linking suggestion. For example, I can still get linking suggestions if e.g. I use a word twice in the meta description, but I have a completely empty post. We should consider implementing a minimum post length for the generation of linking suggestions.

hansjovis:
> I've added an in-line comment to motivate the 4

Thanks!

> We should consider implementing a minimum post length for the generation of linking suggestions.

I agree. I think adding a minimum post length would help in generating relevant suggestions, since it is easier to know what a post is about when there is more content.

PR #349, opened 2019-09-02
Add an Alert component

igorschoester:
## Summary

* [@yoast/components] Adds an Alert component.
* [@yoast/style-guide] Adds new status colors: error, info, success, warning.

* Based on the [MyYoast implementation](https://github.com/Yoast/my-yoast/blob/7a3c8a6cb12cf60bfaad3dc0c1c612e43f4d59dc/client/src/components/general/Alert.js). Should be similar in usage. The cookie and dismiss should now be controlled from outside this component.
* Adds the colors to the style-guide in a hopefully reusable fashion.

## Test instructions

* Run the components app.
* Check the different type of alerts in the components 'tab'.
* Check the dismissable alert.
* The should look like [the design](https://github.com/Yoast/design/issues/371).
* Check in Right To Left: spacing of the icons on the correct sides.

## Impact check

  *

## Screenshot
alerts

## UI changes

## Quality assurance

Partially fixes: https://github.com/Yoast/wordpress-seo/issues/13173


hwinne:
CR Done 👍

packages/components/src/Alert.js
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * Renders the component.
+	 *
+	 * @returns {React.Element} The rendered component.
+	 */
+	render() {
+		const options = this.getTypeDisplayOptions( this.props.type );
+
+		return <AlertContainer color={ options.color } background={ options.background }>
```
afercia: The `color` prop is rendered as an HTML attribute and should be renamed
`<div color="#00468f" class="Alert__AlertContainer ...`

packages/components/src/Alert.js
```diff
@@ -0,0 +1,143 @@
...
+			<AlertIcon icon={ options.icon } color={ options.color } />
+			<AlertContent>{ this.props.children }</AlertContent>
+			{
+				typeof this.props.onDismissed === "function"
+					? (
+						<AlertDismiss
+							color={ options.color }
```
afercia: The `color` prop is rendered as an HTML attribute and should be renamed
`<button color="#00468f"  ...`

packages/components/src/Alert.js
```diff
@@ -0,0 +1,143 @@
...
+Alert.propTypes = {
+	children: PropTypes.any.isRequired,
+	type: PropTypes.oneOf( [ "error", "info", "success", "warning" ] ).isRequired,
+	onDismissed: PropTypes.func,
+	dismissAriaLabel: PropTypes.string,
+};
+
+Alert.defaultProps = {
+	onDismissed: null,
+	dismissAriaLabel: "Dismiss this alert",
```
afercia: I seem to recall passing translatable strings as default didn't work properly for translation, will double check.

packages/components/src/Alert.js
```diff
@@ -0,0 +1,143 @@
...
+	
+	// Override the base button style: get rid of the button styling.
+	min-height: auto;
+	padding: 0;
+
+	&, &:hover, &:focus, &:active {
+		border-radius: 0;
+		border: none;
+		background: transparent;
+		box-shadow: none;
```
afercia: This way, the button has no focus style. On the base buttons, focus style is made via border and box-shadow.

afercia:
About the translatable string in the default prop, see for example https://github.com/Yoast/wordpress-seo/pull/11705#issuecomment-442770156

Seems the default needs to be provided within the render.

packages/components/src/Alert.js
igorschoester: My bad, it did not even register in my brain. Based tests on this too 😞

packages/components/src/Alert.js
igorschoester: Right, the parser is not that smart of course.

packages/components/src/Alert.js
igorschoester: Another good point. Not sure, maybe not override the default one?

packages/components/src/Alert.js
```diff
@@ -0,0 +1,159 @@
...
+import { colors } from "@yoast/style-guide";
+
+/* Internal dependencies */
+import Button from "./buttons/Button";
+import SvgIcon from "./SvgIcon";
+
+const AlertContainer = styled.div`
+	display: flex;
+	align-items: flex-start;
+	fontSize: 14px;
```
afercia: Typo: should be `font-size`

afercia:
I made several changes so I think it's better someone else does final acceptance. Will move back to the proper lane.

Screenshots:

Single line

Screenshot 2019-09-09 at 16 00 51

Multiple lines and focus style for the dismiss button. 

Screenshot 2019-09-09 at 15 53 41

This is based on the provided design on https://github.com/Yoast/design/issues/392#issuecomment-515051908
- 2px border `#007FFF` 
- 3px outer border `#007FFF` with an opacity of 0.25

Note: It is not possible to make the dismiss button alignment have a pixel-perfect alignment because:
- alignment with the single line text matters
- the `&times;` character changes depending on the font used and the font internal metrics. See for example in the plugin, where there's a different font-family thus the alignment is different.

[Noted on the design issue](https://github.com/Yoast/design/issues/371#issuecomment-529504901), pending feedback.

In the plugin:

Screenshot 2019-09-09 at 16 14 56

Screenshot 2019-09-09 at 16 15 22

Screenshot 2019-09-09 at 16 17 32

In the last screenshot you can clearly see the vertical alignment is different. The only reason for this is the font face internal metrics. As noted in the design issue, the only option I can think of to avoid this is to use an SVG icon.

afercia:
Re the `&times;` the decision is to use the current SVG icon, see feedback on https://github.com/Yoast/design/issues/371#issuecomment-529517547

packages/components/src/Alert.js
johannadevos: @igorschoester I don't see Andrea's suggestion implemented in the code?

johannadevos:
Acceptance:  👍

PR #353, opened 2019-09-05
Provide singular and plural form for the Prominent Words occurrences string

### LINKED ISSUE #352: The WordOccurrences component should use `_n()` for proper translation
## Explanation
The WordOccurrences component uses a translatable string that doesn't take into account singular / plural forms:

```
sprintf( __( "%d occurrences", "yoast-components" ), occurrence )
```

Instead, it should use `_n()` to provide singular / plural strings, taking into account whether there's just 1 occurrence or more occurrences:

```
sprintf(
	/* translators: %d: Prominent words occurrences. */
	_n( "%d occurrence", "%d occurrences", occurrence, "yoast-components" ),
	occurrence
)
```

To reproduce
- go to the standalone app demo > WordOccurrences tab
- change one of the words occurrences to `1`
- inspect the source in your dev tools
- see the string is incorrect: `1 occurrences`

afercia:
## Summary

* [components] Improved Prominent Words occurrences strings translation.

- uses `_n()` to provide singular and plural forms
- in the demo app, makes sure the value that comes from the occurrences number field is of type `number`: typically, browsers send values from input fields as `string` even if the input field is of type "number". This was causing an "invalid prop type" warning _and_ the string to not be updated even before this PR (see attached screenshot)
- reduces the ESLint max-warnings to 94

Screenshot 2019-09-05 at 12 57 37

## Test instructions
- go to the standalone demo app > WordOccurrences tab
- inspect the source of one of the words in the list, for example: "Linking"

Screenshot 2019-09-05 at 14 19 40

- see the visually hidden string is `2 occurrences`
- in the related input fields above the list, change the occurrences for this word to `1`
- see the visually hidden string changed to `1 occurrence` (singular)
- repeat on other words with different occurrences values and check there are no errors

## UI changes

## Quality assurance

Fixes https://github.com/Yoast/javascript/issues/352

### DIFF
--- a/apps/components/WordOccurrencesWrapper.js
+++ b/apps/components/WordOccurrencesWrapper.js
@@ -108,7 +108,7 @@ class WordOccurrencesWrapper extends React.Component {
 					nextState.relevantWords[ index ]._stem = input;
 					break;
 				case "_occurrences":
-					nextState.relevantWords[ index ].setOccurrences( input );
+					nextState.relevantWords[ index ].setOccurrences( parseInt( input, 10 ) );
 					break;
 			}
 			return nextState;
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -13,7 +13,7 @@
   "license": "GPL-3.0",
   "scripts": {
     "test": "jest",
-    "lint": "eslint . --max-warnings=96",
+    "lint": "eslint . --max-warnings=94",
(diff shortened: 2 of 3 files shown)


afercia:
After some investigation, decided not to port this PR to the Free repo. Seems to me there's no need to provide a singular form for the number of occurrences as words with just one occurrence won't be considered "prominent". So the current `sprintf( __( "%d occurrences", "yoast-components" ), props.number );` seems fine. The problem was more evident in the demo app, where an arbitrary number of occurrences could be set for testing purposes.
Closing.

PR #356, opened 2019-09-17
Add suffixes to make Dutch nouns

### LINKED ISSUE #355: Add Dutch noun suffixes
PRs:
https://github.com/Yoast/javascript/pull/356
https://github.com/Yoast/YoastSEO.js-premium-configuration/pull/58

We should attach noun suffixes (plural and diminutive) to stems of Dutch words in order to create noun forms. E.g., we want to create the forms _katten_ and _katje_ when the stem is _kat_. This may also involve modifying the stem before adding suffixes (e.g. the _t_ in _kat_ should be doubled before adding the suffix -_en_)

agnieszkaszuba:
## Summary

* Creates forms of regular Dutch nouns by adding suffixes to the stem(s).

## Test instructions

Add more stems of regular nouns and the expected forms to the specs. The following should be tested:
- the suffixes -es and -ers are always added
- the right suffixes are added to predictable stem endings (those stem endings and their suffixes can be found in the morphology data file under "predictedBasedOnStem", both for the diminutive and the plural suffixes)
- if the noun does not end in one of the predictable stem endings, the plural suffix -en is added and the diminutive suffixes -je, -etje, -ertje are added
- before adding the suffixes -en, -es, -ers, -etje and -ertje, stem changes are applied if needed. These stem changes, together with stem endings that they should be applied to, are found in the morphology data file under 'addSuffixes' -> 'stemModifications' (first regex is the stem ending that should be modified, the second regex is what the ending should be replaced with). 
- when the stem ends in -ing, the _g_ should be removed before adding the diminutive suffix _kje_.

There are also nouns which will not get (all) of their correct forms. They are various exceptions to the rules, or unusual nouns such as acronyms or loanwords. These are listed in the morphology data file under "nouns" -> "exceptions".

## Quality assurance

Fixes #355


packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Checks whether the stem has an ending for which the final consonant should not be voiced.
+ *
+ * @param {string} stemmedWord  The stem.
+ * @param {string[]} stemEndings The endings to search for in the stem.
+ * @returns {boolean} Whether the stem has one of the endings that were searched for.
+ */
+const shouldConsonantBeVoiced = function( stemmedWord, stemEndings ) {
+	return stemmedWord.search( new RegExp( stemEndings[ 0 ] ) === -1  ) && stemmedWord.search( new RegExp( stemEndings[ 1 ] ) === -1 );
```
hansjovis: Do stem endings always contain two parts?
If not, it may be worthwhile to loop over the stem endings instead.

packages/yoastseo/spec/morphology/dutch/addNounSuffixesSpec.js
```diff
@@ -0,0 +1,120 @@
+import { getSuffixes } from "../../../src/morphology/dutch/addNounSuffixes";
```
maartenleenders:
```suggestion
import { getSuffixes, addNounSuffixes } from "../../../src/morphology/dutch/addNounSuffixes";
```

packages/yoastseo/spec/morphology/dutch/addNounSuffixesSpec.js
```diff
@@ -0,0 +1,120 @@
...
+import { addNounSuffixes } from "../../../src/morphology/dutch/addNounSuffixes";
```
maartenleenders:
```suggestion
```

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
+import { modifyStem } from "../morphoHelpers/suffixHelpers";
```
maartenleenders:
```suggestion
```

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
...
+import { applySuffixesToStem } from "../morphoHelpers/suffixHelpers";
```
maartenleenders:
```suggestion
import { modifyStem, applySuffixesToStem } from "../morphoHelpers/suffixHelpers";
```

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
maartenleenders:
```suggestion
```

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
maartenleenders:
```suggestion
import { modifyStem, applySuffixesToStem } from "../morphoHelpers/suffixHelpers";
```

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Searches for a match at the end of the stem, and returns the corresponding suffix if matched.
+ *
+ * @param {string} stemmedWord The stemmed word
+ * @param {string[]} suffixType Whether plural or diminutive suffixes are being searched for.
+ * @returns {string} The suffix.
+ *
+ */
+export function findSuffixes( stemmedWord, suffixType ) {
```
maartenleenders: Seems like this function should be called `findSuffix` rather than `findSuffixes` to avoid confusion over the thing it returns.

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
+import { modifyStem } from "../morphoHelpers/suffixHelpers";
+import { applySuffixesToStem } from "../morphoHelpers/suffixHelpers";
+
+/**
+ * Searches for a match at the end of the stem, and returns the corresponding suffix if matched.
+ *
+ * @param {string} stemmedWord The stemmed word
+ * @param {string[]} suffixType Whether plural or diminutive suffixes are being searched for.
```
maartenleenders: These are actually arrays of suffixes, and they can be plural ones or diminutive ones. So `suffixType` is a slightly misleading parameter name. You could just rename this to `stemSuffixPairs`, or simply `suffixes`.

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
...
+ * @returns {string} The suffix.
+ *
+ */
+export function findSuffixes( stemmedWord, suffixType ) {
+	const stemEndingAndSuffix = suffixType.find( stemSuffixPair => stemmedWord.search( new RegExp( stemSuffixPair[ 0 ] ) ) !== -1 );
+
+	if ( typeof stemEndingAndSuffix !== "undefined" ) {
```
maartenleenders: It is more common to just check whether `stemEndingAndSuffix` is "truthy":
```suggestion
	if ( stemEndingAndSuffix ) {
```

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Gets the diminutive or plural suffix predicted based on stem ending or the default suffix(es) if no suffix can be predicted.
+ *
+ * @param {string} stemmedWord The stemmed word.
+ * @param {Object} morphologyDataNounSuffixes The morphology data with either plural or diminutive noun suffixes.
+ * @returns {Array} The suffixes.
+ */
+export function getSuffixes( stemmedWord, morphologyDataNounSuffixes ) {
+	const suffixes = [];
```
maartenleenders: This suffixes variable is only needed inside the `if` loop below.

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
...
+	const predictedSuffix = findSuffixes( stemmedWord, morphologyDataNounSuffixes.predictedBasedOnStem );
+
+	if ( typeof predictedSuffix !== "undefined" ) {
+		suffixes.push( predictedSuffix );
```
maartenleenders: you can 
`return [ predictedSuffix ];`
immediately, without first creating an empty array and pushing it in.

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @param {string} stemmedWord The stemmed word.
+ * @param {Object} morphologyDataNounSuffixes The morphology data with either plural or diminutive noun suffixes.
+ * @returns {Array} The suffixes.
+ */
+export function getSuffixes( stemmedWord, morphologyDataNounSuffixes ) {
+	const suffixes = [];
+	const predictedSuffix = findSuffixes( stemmedWord, morphologyDataNounSuffixes.predictedBasedOnStem );
+
+	if ( typeof predictedSuffix !== "undefined" ) {
```
maartenleenders:
```suggestion
	if ( predictedSuffix ) {
```

See above

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
+import { modifyStem } from "../morphoHelpers/suffixHelpers";
+import { applySuffixesToStem } from "../morphoHelpers/suffixHelpers";
+
+/**
+ * Searches for a match at the end of the stem, and returns the corresponding suffix if matched.
+ *
+ * @param {string} stemmedWord The stemmed word
+ * @param {string[]} suffixType Whether plural or diminutive suffixes are being searched for.
+ * @returns {string} The suffix.
```
maartenleenders: this can also return `void`, perhaps you can use `{string|void} The suffix, if one was found`, or something like that, I am unsure myself.

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,170 @@
...
+
+	const extraPluralSuffixes = findSuffixes( stemmedWord, morphologyDataPluralSuffixes.extraSuffixes );
+
+	if ( typeof extraPluralSuffixes !== "undefined" ) {
+		pluralSuffixes = pluralSuffixes.concat( extraPluralSuffixes );
+	} return pluralSuffixes;
+};
+
+/**
+ *
```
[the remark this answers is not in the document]
maartenleenders: Additionally, for me it was unclear what the purpose of the function is.

I mean, I can somewhat see that it indeed
>Searches for a given suffix in an array, and if found, moves it to another array.

But I am not sure why we are splicing suffixes from one array, and pushing them to another.

If you also describe the spirit and purpose of the function, coding becomes easier for others.

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
...
+
+	const suffixesForModifiedStem = [];
+	const kjeSuffix = [];
+
+	// Make extra suffixes groups for suffixes that require a stem modification.
+	moveSuffixes( "en", pluralSuffixes, suffixesForModifiedStem );
+	moveSuffixes( "ers", pluralSuffixes, suffixesForModifiedStem );
+	moveSuffixes( "es", pluralSuffixes, suffixesForModifiedStem );
+	moveSuffixes( "etje", diminutiveSuffixes, suffixesForModifiedStem );
+	moveSuffixes( "kje", diminutiveSuffixes, kjeSuffix );
```
maartenleenders: This part is a bit serial for what is actually happening, and also perhaps a bit unclear. 

As you told to me, the goal here is to move any `"en", "ers", "es"` from the `pluralSuffixes`, and any `"etje"` from `diminutiveSuffixes`, to the `suffixesForModifiedStem` because they require modification of the stem. Additionally, `"kje"` should be moved from diminutive to its own array if present.

So I actually suggest a bit of a rewrite for this part. I have an idea, but it is a bit too much to explain in a CR comment, so would be good to pair on it (or alternatively I can push the changes myself).

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
```diff
@@ -0,0 +1,172 @@
...
+ * @param {array} oldGroup The group in which the suffix should be searched for.
+ * @param {array} newGroup The group that the suffix should be moved to if found.
+ * @returns {void}
+ */
+const moveSuffixes = function( suffix, oldGroup, newGroup ) {
+	const suffixIndex = oldGroup.indexOf( suffix );
+
+	if ( suffixIndex !== -1 ) {
+		oldGroup.splice( suffixIndex, 1 );
+		newGroup.push( suffix );
```
maartenleenders: In general it's not so nice to change objects that are passed to the function by reference (as you do here: oldGroup is changed by the splice, so anyone calling `moveSuffixes( "a", myArray, newArray )` will change both myArray and newArray by doing so, which can lead to unclear situations.

Now, in this specific case, it is probably fine, as the whole purpose of the function is to alter the arrays, I just wanted to mention that it's usually avoided 😉 .

That being said, I think this function is unnecessary (see below) and it's goal can be achieved differently. I'll submit some code to that end.

packages/yoastseo/src/morphology/dutch/addVerbSuffixes.js
```diff
@@ -27,8 +28,8 @@ const findAndApplyModifications = function( stemmedWord, morphologyDataAddSuffix
 	if ( triedToDoubleConsonant ) {
 		return triedToDoubleConsonant;
 	}
-	if ( shouldConsonantBeVoiced( stemmedWord, morphologyDataAddSuffixes.otherChecks.noConsonantVoicingVerbs ) ) {
-		const triedToVoiceConsonant = modifyStem( stemmedWord, morphologyDataAddSuffixes.stemModifications.consonantVoicing );
+	if ( shouldConsonantBeVoiced( stemmedWord, morphologyDataAddSuffixes.otherChecks.noConsonantVoicingNounsVerbs ) ) {
```
maartenleenders: If I am not mistaken, this `findAndApplyModifications` function is actually identical to the one in `addNounSuffixes`.

I suggest adding it to the `suffixHelpers` and importing it in both `addVerbSuffixes` and `addNounSuffixes`

packages/yoastseo/src/morphology/dutch/addVerbSuffixes.js
```diff
@@ -6,11 +6,12 @@ import { applySuffixesToStem } from "../morphoHelpers/suffixHelpers";
  * Checks whether the stem has an ending for which the final consonant should not be voiced.
  *
  * @param {string} stemmedWord  The stem.
- * @param {string[]} stemEndings The endings to search for in the stem.
+ * @param {string[]} notVoicedStemEndings The endings to search for in the stem.
  * @returns {boolean} Whether the stem has one of the endings that were searched for.
  */
-const shouldConsonantBeVoiced = function( stemmedWord, stemEndings ) {
-	return stemmedWord.search( new RegExp( stemEndings[ 0 ] ) === -1  ) && stemmedWord.search( new RegExp( stemEndings[ 1 ] ) === -1 );
+const shouldConsonantBeVoiced = function( stemmedWord, notVoicedStemEndings ) {
```
maartenleenders: this is actually identical to the one in `addNounSuffixes`, I suggest moving it to the `suffixHelpers` and importing in `addVerbSuffixes` and `addNounSuffixes`.

packages/yoastseo/src/morphology/dutch/addVerbSuffixes.js
maartenleenders: Note that importing `findAndApplyModifications` should then be done from the suffixhelpers, not the one that is exported from `addAdjectiveSuffixes`, which is slightly different internally. I suggest renaming that function to `findAndApplyAdjectiveModifications` for clarity

packages/yoastseo/src/morphology/dutch/addNounSuffixes.js
maartenleenders: Added the code in https://github.com/Yoast/javascript/pull/356/commits/35c3841b90e87de5f8740e134730bf2a5f7d555f

packages/yoastseo/src/morphology/dutch/addVerbSuffixes.js
maartenleenders: Ah, I have now discovered that the `suffixHelpers` is language-agnostic and these functions are language specific.

In that case you can disregard my comment above, but perhaps do consider making language specific suffixHelpers. Like a `morphology/dutch/dutchSuffixHelpers.js` for functions that are used for all Dutch suffixes (irrespective of noun or verb (or adjective, though those seem even more different)).

I'd be happy to discuss.

packages/yoastseo/src/morphology/dutch/addVerbSuffixes.js
maartenleenders: Ah, I've since learned that my recommendation is not completely correct. So disregard, and see:
https://github.com/Yoast/javascript/pull/356#discussion_r349530315

maartenleenders:
CR: needs a bit more clarity to keep the codebase legible and keep the threshold for maintenance low.

I've also changed some of the JavaScript logic, so please review those changes if you are interested. If you have any questions, feel free to ask me.

I've made a recommendation about `dutchSuffixHelpers`, but I don't know enough about the ideal intended architecture for your usual work. I'd be happy to discuss.

maartenleenders:
CR 👍

maartenleenders:
Acceptance: LGTM 👍

PR #377, opened 2019-10-16
Adds exceptions (indeclinable and exceptions with full forms)

### LINKED ISSUE #371: Exception words with full forms
Adds and unites all exceptions and their forms.

hannaw93:
## Summary

Checks whether a word is one of the exceptions for which we list all of the forms and the canonical stem. If it matches one of the listed forms, only return the stem .

* The code is made as a morphoHelper function as it can be reused later for different language. 

## Test instructions

* Add more words in spec file. 
The words can be taken from stemmingExceptionStemsWithFullForms list in the morphology data file. E.g. "zijden" from this list.
["zij", ["zij", "zijde", "zijden", "zijdes", "zitje", "zijtjes"]]
The expected result of the test: the stem (i.e. "zij")

## Quality assurance

Fixes #371 
Fixes https://yoast.atlassian.net/browse/LIN-103


packages/yoastseo/src/morphology/morphoHelpers/checkExceptionsWithFullForms.js
```diff
@@ -0,0 +1,34 @@
...
+		const fullFormsExceptionList = flatten( fullFormsExceptionLists[ i ] );
+		for ( let j = 0; j < fullFormsExceptionList.length; j++ ) {
+			if ( word.endsWith( fullFormsExceptionList[ j ] ) ) {
+				/*
+				 * Check if the word checked is actually a compound word. e.g. familielid
+				 * The character/s preceding the words in the exception should be at least 2 characters in order to be a valid compound element.
+				 */
+				const precedingLexicalMaterial = word.slice( 0, -fullFormsExceptionList[ j ].length );
+				if ( precedingLexicalMaterial.length === 1 ||
+					 precedingLexicalMaterial.length === 2 ) {
```
agnieszkaszuba: I think the `precedingLexicalMaterial.length === 2' shouldn't be there? The preceding lexical material should be AT LEAST 2 characters, so larger than 1.

packages/yoastseo/src/morphology/morphoHelpers/checkExceptionsWithFullForms.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ *
+ * Checks whether a word is on the exception list for which we have full forms.
+ *
+ * @param {Array} fullFormsExceptionLists The exception list with full forms to check against.
+ * @param {string} word The word to check.
+ *
+ * @returns {string/null} The created word forms.
+ */
+export default function( fullFormsExceptionLists, word ) {
```
agnieszkaszuba: I would change the naming of the variables `fullFormExceptionLists` and `fullFormExceptionList` to make it more clear that each 'list' it loops through is one word and its forms and stem. Maybe: 'fullFormExceptionList' (instead of Lists) and 'stemAndForms'?

packages/yoastseo/src/morphology/morphoHelpers/checkExceptionsWithFullForms.js
moorscode: Isn't that the condition below?

agnieszkaszuba: We need a different outcome for when precedingLexicalMaterial = 1 compared to when it equals 0. So the conditions below are not enough. We need this extra line of code to specify what happens when precedingLexicalMaterial = 1.

PR #404, opened 2019-11-19
2547 make prominent word naming clearer

hansjovis:
## Summary

* Refactors the code to use another term than 'prominent words', 'keyword suggestions' or 'relevant words'.

* I boyscouted the `WordOccurrencesWrapper` class in the components app. I added comments and renamed some variables, methods and functions.

## Test instructions

**Note**: this PR needs to be tested in unison with the premium PR: https://github.com/Yoast/wordpress-seo-premium/pull/2613

* Check if the readability- and SEO-analysis work as expected.
* Check if the word occurrences insight works as expected.
* Check if the internal linking suggestions work as expected.
* Check if site-wide internal linking works as expected.

## Impact check

  *

## UI changes

## Quality assurance

Fixes #https://github.com/Yoast/wordpress-seo-premium/issues/2547


apps/components/WordOccurrencesWrapper.js
```diff
@@ -27,136 +28,242 @@ const initialRelevantWords = [
...
 
 		this.state = {
-			relevantWords: cloneWords( initialRelevantWords ),
+			words: cloneWords( initialRelevantWords ),
 		};
-		this.changeRelevantWord = this.changeRelevantWord.bind( this );
-		this.removeRelevantWord = this.removeRelevantWord.bind( this );
-		this.addRelevantWordRow = this.addRelevantWordRow.bind( this );
-		this.resetRelevantWords = this.resetRelevantWords.bind( this );
+		this.change = this.change.bind( this );
```
maartenleenders: In my opinion this refactor has gone too far in the other direction: `this.change()` is essentially `WordOccurrencesWrapper.change()` which does not tell me what I am doing. `changeWord`,`removeWord`, `addWord` would have been fine for the issue, and more descriptive than this. 

Then again it's just the test app so 🤷‍♀

packages/components/src/index.js
```diff
@@ -43,7 +43,8 @@ export { default as Icon } from "./Icon";
 export { default as IconButtonToggle } from "./IconButtonToggle.js";
 export { default as IFrame } from "./IFrame";
 export { default as Input } from "./input/Input";
-export { default as KeywordSuggestions } from "./KeywordSuggestions";
+export { default as KeywordSuggestions } from "./WordOccurrenceInsights";
```
nataliashitova: I understand that it is important that apps that make use of `KeywordSuggestions` don't break in this commit. But in a log run, we don't want things like that in our code. 
@atimmer Is it possible to throw a deprecation warning in this situation?

packages/components/src/WordOccurrenceInsights.js
```diff
@@ -64,24 +64,24 @@ const getKeywordSuggestionExplanation = keywords => {
  * @summary WordList component.
  *
  * @param {string}   title           The title of the list.
- * @param {WordCombination[]|ProminentWord[]}   relevantWords   The relevant words.
+ * @param {WordCombination[]|ProminentWord[]}   words   The relevant words.
```
nataliashitova:
```suggestion
 * @param {WordCombination[]|ProminentWord[]}   words   The words to display in the list.
```

packages/components/src/index.js
atimmer: It's possible to throw a deprecation warning, but in this case I would:

* Move this line to the bottom of the file with a `// Deprecated:` comment above it.
* Add `export { default as KeywordSuggestions } from "./WordOccurrenceInsights";` as the new way to import the component in the future.
* Add a line to the changelog detailing that `KeywordSuggestions` is deprecated and should be replaced with `WordOccurrenceInsights`.

hansjovis: I implemented the suggested changes 🙂.

PR #422, opened 2019-12-11
Add window.location.origin as second argument to the URL constructor

### LINKED ISSUE #418: Creating the web worker fails on relative URLs
## Explanation
When a website uses relative URLs to point to assets, e.g. `/worker-script.js`, creating the web worker fails and the SEO- and Readability Analysis cannot be run.

## Technical details
When creating the web worker we check whether the URL to the worker script
(e.g. `https://example.org/worker-script.js` when calling:
```js
const webworker = new Worker( 'https://example.org/worker-script.js' );
```
) has the same origin as the current website (e.g. `https://some-other-website.org`).

We check this by casting both the website as well as the path to the worker to `URL` objects and checking their host names.

This fails when a website uses relative URLs, e.g. `/worker-script.js`, instead of the above absolute URLs. Since a relative URL cannot be cast to `URL` objects:
```js
// This fails:
const url = new URL( '/worker-script.js' );
```

## Solution
The same origin check is done here:

https://github.com/Yoast/javascript/blob/7bbe79da581e30d88e8b38eb63fa6bb01745c81a/packages/yoastseo/src/worker/createWorker.js#L39-L46

We need to check whether `urlA` or `urlB` are relative URLs and, if so, return `true`. Relative URLs are relative to the current origin, so are necessarily from the same origin.

igorschoester:
This will add support to relative as well as absolute URLs.

## Summary

* [yoastseo] Fixes a bug where the metabox would be broken when a relative URL was configured as `WP_CONTENT_URL`. Props to [FPCSJames](https://github.com/FPCSJames).

*

## Test instructions

### To reproduce
* Run/activate the Yoast SEO's current release.
* Add `define('WP_CONTENT_URL', '/wp-content');` to for example `wp-config.php`.
* Go to a post.
* It is expected that you get the following error in the console: `createWorker.js?a02a:41 Uncaught TypeError: Failed to construct 'URL': Invalid URL`

### Test PR
* Switch to this branch.
* Verify the error is no longer there.
* Test the analysis results (SEO & Readability). Basically it is already working because of no error, but if you can get a point to change it definitely communicated correctly.
* Remove the `define('WP_CONTENT_URL', '/wp-content');` you entered in the reproduce steps.
* Verify the error is still not there.
* Test the analysis results again.
* Add `define('WP_CONTENT_URL', 'http://local.wordpress.test/wp-content');` to for example `wp-config.php`. If your test site URL is different, please adjust it accordingly. The point is to test an absolute URL override.
* Verify the error is still not there.
* Test the analysis results again.

## Impact check

  *

## UI changes

## Quality assurance

Fixes #418

### DIFF
--- a/packages/yoastseo/src/worker/createWorker.js
+++ b/packages/yoastseo/src/worker/createWorker.js
@@ -37,8 +37,8 @@ function createBlobScript( url ) {
  * @returns {boolean} Whether the URLs have the same origin.
  */
 function isSameOrigin( urlA, urlB ) {
-	urlA = new URL( urlA );
-	urlB = new URL( urlB );
+	urlA = new URL( urlA, window.location.origin );
+	urlB = new URL( urlB, window.location.origin );
 
 	return urlA.hostname === urlB.hostname &&
 		urlA.port === urlB.port &&


Djennez:
Acceptance done 👍

PR #434, opened 2020-01-08
Add include merged tags option to the lerna config

igorschoester:
This is needed because the "RC" tags are created on release branches and the "release" tags on the master branch.

## Summary

* Fixes a bug where the `lerna version` command would want to publish seemingly random packages.

* `lerna version` uses `lerna changed`, which uses `git describe --always --long --dirty --first-parent` by default. It has the option `--include-merged-tags` that removes the `--first-parent` from the command, so that makes it easier to test this.

## Test instructions

In order to make it easier for me to give you the same results I'm seeing, I'm writing these instructions with the current situation, the RCs for the 12.9 release.

See what Lerna uses as base-tag by running the changed command yourself:
* Go to the commit before the RC commit: `git checkout cf24c03c4bda64013536adf67735e3c64056b26c`.
* Run `yarn lerna changed`:
* Verify the output has a line that says: `lerna info Looking for changed packages since @yoast/analysis-report@0.12.0-rc.0`.
  * The last part is the tag name. Which is unexpected and wrong.
* Run `yarn lerna changed --include-merged-tags`
* Verify the `changed packages since` is now `@yoast/analysis-report@0.12.0`.
  * This verifies that it uses the correct base tag, let's continue with the packages.
* Verify the output of the `yarn lerna change --include-merged-tags` has the following 5 packages:
  * @yoast/analysis-report
  * @yoast/components
  * @yoast/configuration-wizard
  * @yoast/search-metadata-previews
  * yoast-components
* Run `git diff --name-only master`.
* Verify that the actual changes between master and this commit only include 2 packages: `components` and `search-metadata-previews`.
  * Both of these are in the list of changed packages.
* Verify that the other 3 packages have either `components` as dependency or `search-metadata-previews`:
  * @yoast/analysis-report has components as a dependency.
  * @yoast/configuration-wizard has components as a dependency.
  * yoast-components has components as a dependency.

The actual code change is the way to configure Lerna to use the `--include-merged-tags` flag. You could (sort of) verify this by running `yarn lerna changed --include-merged-tags` and `yarn lerna changed` on this branch and see that there is no difference.
As for `lerna changed` being used by `lerna version`... I did not dare to run the actual version command because it will publish. But you could look at their documentation, the code, or actually implement a break in the code to verify this does work.

## Impact check

  * No actual release is being done here, but it should influence those.

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,6 @@
 {
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "independent"
+  "version": "independent",
+  "includeMergedTags": true
 }


igorschoester:
I guestimate 8 points

hansjovis:
Acceptance test: ✅ 

The `lerna changed` command works as expected and described in the test instructions now.

PR #458, opened 2020-01-21
Merge pull request #458 from Yoast/435-dutch-participle-stemming-exceptions

### LINKED ISSUE #435: Add exceptions and irregular cases to stemming Dutch participles

agnieszkaszuba:
## Summary

* Adds exceptions to the stemming of the participles of regular verbs.

* Take more participles from the `doNotStemGe` list in the morphology data file, add them to the specs, and make sure that the ge- is not stemmed. Also try adding separable prefixes to the participles and ensure that the ge- is still not stemmed. For example, make sure that ge- is not stemmed both in _gebruikt_ and in _opgebruikt_.
* Make sure that participles with the prefixes _herin_, _herop_, and _heruit_ (e.g. heropgebouwd -> heropbouw) are stemmed correctly. Also that the participles with the prefix -her, not followed by -in/op/uit are stemmed correctly (e.g. herhaald -> herhaal).

## Relevant technical choices
The name of an exception list in the premium configuration data file was changed to the more clear inseparableCompoundVerbs. The name of a similar list for _generating_ participle forms, was changed to match this one, to inseparableCompoundVerbsStems. As a result, the reference to the data object was changed in a file for creating participle forms (generateParticipleForms).

Fixes #435
Fixes https://yoast.atlassian.net/browse/LIN-102?atlOrigin=eyJpIjoiYWVlNzY4YzdkZGMzNGYyOTk4YzA4YTgzYTBjMWQxNzMiLCJwIjoiaiJ9


nataliashitova:
I wonder what the reasons were why only three compound prefixes with _her_ were added. I can imagine that any separable or inseparable prefix can be used in combination with _her_ (because of its very general meaning), e.g. `heraangebracht` or `heringebracht`. Are the three you added more often used in the language?

nataliashitova:
From @agnieszkaszuba and @hannaw93 I understood that a 250,000 word corpus was used to assess which prefixes should be added to the rule and that only those added in the current PR were found in the corpus. This means that other combinations are (even) more rare. For the sake of pragmatism, we will go ahead with this solution and re-evaluate it if needed at a later time point.

PR #497, opened 2020-02-11
Check whether the -t/d at the end of a detected participle should be stemmed or not

agnieszkaszuba:
## Summary

* Adds a check to the Dutch participle stemmer to make sure that when word-final -t/-d is part of the stem, it does not get stemmed.

* We use an exception list of verbs ending in -t that shouldn't be stemmed for one of the checks. This list is part of the larger 'wordsNotToBeStemmed' object that is checked at the beginning of the non-participle stemmer. The verbs ending in -t were separated from the other words and listed as a separate array inside of the object. Because of this new nested structure, the object is now flattened before checking the list in stem.js

* The list of verbs ending in -t that shouldn't be stemmed was also expanded with some additional entries. The word 'eest' was added, which caused the adjective 'tevreest' to not be stemmed anymore in stemSpec.js. This is because we check whether the current word ends with an entry from the exception list of words that shouldn't be stemmed, and 'tevreest' ends with 'eest'. This will be fixed after dividing the exception lists into those where we match with the ending of the word, and those where we look for the exact match. This will be a separate issue. For now, 'tevreest' was removed from stemSpec.js to avoid failed tests.

## Test instructions

* Add more examples of participles that fall under the added exception checks to the specs, and make sure that they are stemmed correctly.

## Quality assurance

### DIFF
--- a/packages/yoastseo/spec/morphology/dutch/detectAndStemRegularParticipleSpec.js
+++ b/packages/yoastseo/spec/morphology/dutch/detectAndStemRegularParticipleSpec.js
@@ -6,37 +6,61 @@ const morphologyDataNL = getMorphologyData( "nl" ).nl;
 
 describe( "Detects and stems participles", () => {
 	it( "does not stem a word if it is on an exception list of non-participles", () => {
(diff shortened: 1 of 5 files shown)


manuelaugustin:
CR & acceptance 👍

PR #498, opened 2020-02-12
Adds morphology data check in retrieveStemmer

manuelaugustin:
Checks whether morphology data is available when retrieving language-specific stemmers from the factory instead of executing that check locally in each language-specific stemmer.

## Summary

* [yoastseo] Checks whether morphology data is available when retrieving language-specific stemmers from the factory instead of executing that check locally in each language-specific stemmer.

* When checking whether the identity function is returned in the specs, I do so by actually running it and checking its output. I initially tried to test whether the function that's returned is indeed the identity function, but jest then tells me `Expected value to equal: [Function anonymous] Received: [Function anonymous]` and `Compared values have no visual difference.`. That's why I'm using this workaround now which still tells me that what is returned is indeed the identity function.
* I removed the test for no morphology data in `determineStemEnglishSpec.js` because `determineStem` function shouldn't be called anymore when there is no morphology data.
* I've made an issue to also use `retrieveStemmer` in the internal linking functionality after all the necessary merges: https://yoast.atlassian.net/browse/LIN-262

## Test instructions

* Run the tests and check that everything passes.
* In the codebase, confirm that all occurrences of `retrieveStemmer` now also get the morphology data passed as an argument.
* In the codebase, confirm that `determineStem` is only ever called via `retrieveStemmer`.

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LIN-232

### DIFF
--- a/packages/yoastseo/src/researches/buildTopicStems.js
+++ b/packages/yoastseo/src/researches/buildTopicStems.js
@@ -1,6 +1,5 @@
-import getStemForLanguageFactory from "../helpers/getStemForLanguage.js";
-
 import filterFunctionWordsFromArray from "../helpers/filterFunctionWordsFromArray.js";
+import retrieveStemmer from "../helpers/retrieveStemmer.js";
 import getWords from "../stringProcessing/getWords.js";
 import parseSynonyms from "../stringProcessing/parseSynonyms";
 import { normalizeSingle } from "../stringProcessing/quotes";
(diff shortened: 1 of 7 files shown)


packages/yoastseo/src/researches/buildTopicStems.js
```diff
@@ -81,19 +78,12 @@ const buildStems = function( keyphrase, language, morphologyData ) {
...
-			word => new StemOriginalPair(
-				normalizeSingle( escapeRegExp( word.toLocaleLowerCase( language ) ) ),
-				word
-			),
-		);
 
-		return new TopicPhrase( lowerCasedOriginalPairs );
-	}
+	/**
+	 * Extract a stemming function (it available, and if there is morphologyData available for this language).
```
manuelaugustin:
```suggestion
	 * Extract a stemming function (if available, and if there is morphologyData available for this language).
```

PR #512, opened 2020-02-19
Remove error states from twitter and facebook image components

### LINKED ISSUE #127: Remove error states from the FacebookImage and TwitterImage components
Currently, the FacebookImage component contains two error states: one which is shown when the uploaded image is too small, and one general error message. Those are shown within the FacebookImage component itself:
Schermafbeelding 2019-03-26 om 10 53 50

However, we've decided to move the error messages about images to a notification, see: https://github.com/Yoast/roadmap/issues/21#issuecomment-469585683 

This means the error states need to be removed from the FacebookImage component. Since the TwitterImage component originally was a copy from the FacebookImage component, TwitterImage should also be checked and the relevant error states should be removed.

GuidoOffermans:
## Summary

* Removed the error states.
* Made sure it always displays the placeholder when errored, loading or there is no image.

* Kept the `error state` of the component. So even when something went wrong it will display the placeholder and it doesn't crash.

## Test instructions

* Check if the previews display a placeholder when errored, loading or when there is no image.

## Impact check

  *

## UI changes

## Quality assurance

Fixes #127

### DIFF
--- a/packages/social-metadata-previews/src/facebook/FacebookImage.js
+++ b/packages/social-metadata-previews/src/facebook/FacebookImage.js
@@ -13,9 +13,6 @@ import {
 	FACEBOOK_IMAGE_SIZES,
 } from "../helpers/determineImageProperties";
 
-const MIN_IMAGE_WIDTH = FACEBOOK_IMAGE_SIZES.squareWidth;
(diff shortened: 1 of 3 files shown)


bintzandt:
ACC 👍 

The Facebook and Twitter preview components still look good!

PR #515, opened 2020-02-24
Add modal components

JoseeWouters:
## Summary

* Work in progress

*

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #


packages/components/src/modal/modal.html
```diff
@@ -0,0 +1,379 @@
...
+					</label>
+
+					<label class="yoast-field-group__radiobutton">
+						<input type="radio" name="exclude-news-sitemap" value="no" checked>
+						No
+					</label>
+				</div>
+
+				<div class="yoast-field-group">
+					<label class="yoast-field-group__title" for="foobar">
```
JoseeWouters: Is there a reason you're using "foobar"? Is it meant as a placeholder?

packages/components/src/modal/components/input/input.css
```diff
@@ -0,0 +1,41 @@
...
+
+.yoast-field-group__inputfield + .description,
+.yoast-field-group .description + .yoast-field-group__inputfield{
+    margin-top: 8px;
+    margin-bottom: 24px;
+}
+
+
+/* Placeholder Firefox */
+::placeholder {
```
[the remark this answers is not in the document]
Sjardo: Old habits, used to be like this otherwise it did not work :D

afercia:
I've been asked to audit the HTML for accessibility. So here it is. As it's longish, I thought to use [disclosure widgets](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) using _native HTML_ 🙂 

<details>
<summary>Switching between tabs and accordions</summary>

From an accessibility perspective, I'd say this is the main issue.

On small screens, the tabbed menu on the left changes to accordions. This is fine but it shouldn't be just a visual thing. Tabs and accordions are well defined UI design patterns that need to implement the expected semantics and interaction.

These should be implemented as ARIA tabs and ARIA accordions.

You can find all the details on the ARIA Authoring Practices including the detailed expected interaction, working examples, and code examples: 
https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel
https://www.w3.org/TR/wai-aria-practices-1.1/#accordion

Worth noting that also the markup needs to be different thus the switch can't be just a CSS thing.

Worth reminding we do have a React component for ARIA tabs. We also have a `collapsible` React component (used in the Yoast SEO plugin), which is very close to an ARIA accordions implementation.
</details>

<details>
<summary>Left menu details</summary>

Related to the above: the left "menu" isn't a navigation menu. It doesn't trigger navigation. Instead, it just reveals / hides content on the page. This is why it needs to be implemented as ARIA tabs.

In any case, we should never, ever, introduce links with a `href="#"` attribute. Historically, this pattern was a developers convenience but it's harmful for users.

UI controls that trigger a behavior aren't links. As such, a `<nav>` wrapper would be inappropriate. These are ARIA tabs.
</details>

<details>
<summary>Click event on the headings (for the accordions)</summary>

There's a click event handler bound on the headings. That is: a non-interactive element has interaction.

Events should be set only on native interactive HTML elements or specific ARIA widgets elements when required by that specific ARIA pattern.

This UI needs to be an ARIA accordions interface anyways, where the headings contain a full-width button (the event is on the button). See https://www.w3.org/TR/wai-aria-practices-1.1/#accordion

Aside 1: worth noting the click callback toggles the CSS class also on desktop, where it's unnecessary and may lead to async state when the view switches to responsive.
Aside 2: this implies the whole toggling mechanism based on `:not(:target)` can't be used.
</details>

<details>
<summary>The modal accessibility</summary>

The modal itself will need many accessibility features that aren't included in this PR.
For example: aria roles, proper labelling, focus management, constraining tabbing, hide other content from assistive technologies, the modal main heading should be a h1.

I'm assuming the modal will use our modal React component?

Aside: as noted by JoseeWouters, the modal "X close" button is empty.
</details>

<details>
<summary>The accordions toggle button</summary>

As said above, these should be implemented as ARIA accordions.
- in any case, the term "toggle" e.g. `Toggle section` should be avoided in strings as it's difficult to translate in other languages, see for example https://core.trac.wordpress.org/ticket/34753
- visually, there's only an icon: speech recognition software users need to figure out the control's name even on mobile: if they don't _see_ the control's name they're not able to voice a proper command
- by using ARIA accordions, speech recognition software would be able to just voice e.g. `Click Insights` or `Click Advanced` and that would work
</details>

<details>
<summary>Help links</summary>

Example of current code:
`<a class="yoast-field-group__help" aria-expanded="false" href="#">`

Will this be a real link to an external Help page or a button to reveal inline help?
</details>

<details>
<summary>Form fieldset and legend elements</summary>

As noted by JoseeWouters some form elements need to be grouped in a fieldset.
However, a fieldset just groups the controls. It doesn't give the group any name.
In some cases, depending on the actual form control _labels_, a `<legend>` element is necessary.

For example: two radio buttons with labels "Yes" and "No".
In this case, when tabbing directly to the radio buttons, the screen reader would announce just "Yes" and "No" with no other context.

A `<legend>` element containing the actual question would solve this issue.
See yes/no example and explanation on https://accessibility.blog.gov.uk/2016/07/22/using-the-fieldset-and-legend-elements/

In general, checkboxes and radio buttons always need a fieldset with a legend. Other grouped form controls may need it, this should be evaluated on a case by case basis.

Worth noting in the Yoast SEO plugin we're using fieldset + legend for good reasons, see for example "Should search engines follow links on this post" in the meta box "Advanced" section.

Example:
in `checkbox.html` the bold text "label name" should be a filedset legend and not contain any other content
</details>

<details>
<summary>Form controls labelling</summary>

Always use implicitly associated label with for/id attribubtes. Do not use wrapping labels.
While this is technically valid:
```
	<label class="yoast-field-group__radiobutton">
		<input type="radio" name="exclude-news-sitemap" value="yes">
		Yes
	</label>
```

it is not guaranteed to work with some browser / assistive technologies combinations, especially some old combos.

For this reason, in the Yoast SEO plugin we're using the for/id pattern as much as we can.
</details>

<details>
<summary>Labels and legends should not contain extraneous content e.g. links</summary>

The `<label>` element is exclusively meant to give a form control its name.
Please remove or move out extraneous content from all the `<label>`s.

For example the Help link within the label:
```
<label class="yoast-field-group__title" for="stock-tickers">
	Stock tickers
	<a class="yoast-field-group__help" aria-expanded="false"
	   href="https://yoast.com/link-to-stock-tickers" target="_blank">
<span class="yoast-field-group__help-icon" aria-hidden="true">
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 12"><defs/><path
	d="M12 6A6 6 0 110 6a6 6 0 0112 0zM6.2 2C4.8 2 4 2.5 3.3 3.5l.1.4.8.7.4-.1c.5-.5.8-.9 1.4-.9.5 0 1.1.4 1.1.8s-.3.6-.7.9C5.8 5.6 5 6 5 7c0 .2.2.4.3.4h1.4L7 7c0-.8 2-.8 2-2.6C9 3 7.5 2 6.2 2zM6 8a1.1 1.1 0 100 2.2A1.1 1.1 0 006 8z"/></svg>
</span>
		<span class="screen-reader-text">Link to information about stock tickers.</span>
	</a>
</label>
```

would be announced as `Stock tickers Link to information about stock tickers`
</details>

<details>
<summary>Links that open a new browser's tab</summary>

All the links with `target="_blank"` need to communicate a new tab will open.
In Yoast SEO, we're using a small utility method for this purpose. A JavaScript equivalent exists in our components.
Our standard text is `(Opens in a new browser tab)`.
See for example `WPSEO_Admin_Utils::get_new_tab_message()`
</details>

<details>
<summary>"Setting saved" needs to be announced by screen readers</summary>

"Setting saved" should be an aria-live region or use a `speak()` message
</details>

<details>
<summary>The "Insights" counts</summary>

They will need a way to communicate what those numbers are. This will need some thinking.
</details>

<details>
<summary>General semantics</summary>

Everything is within a `<form>` element.

- from a semantic perspective, the menu (to be changed to ARIA tabs / accordions) is arguably part of the form
- I'm not sure the structure made of `<section>`, `<header>`, `<article>` elements is appropriate within a form
- usually, `<section>`s are within `<article>`s: is there a reason why it's the opposite case here? References:
  - https://www.w3.org/TR/html52/sections.html#the-article-element
  - https://www.w3.org/TR/html52/sections.html#the-section-element
  - https://www.w3.org/wiki/HTML/Usage/Headings/Missing
- not sure the usage of the `footer` element is correct, reference:
  - https://www.w3.org/TR/html52/sections.html#the-footer-element
  - > A footer typically contains information about its section, such as who wrote it, links to related documents, copyright data, and the like.
  - currently, the footer contains "Settings saved" and a button, which seems inappropriate
</details>

<details>
<summary>Descriptions</summary>

See for example in `input.html`

The descriptions need to be programmatically associated with their input field via an `aria-describedby` attribute on the field referencing the id of the description. We're doing this in Yoast SEO already.

Consider to place the description _above_ the field. For xample: browser auto-fill will hide the description.
</details>

Sjardo:
@bintzandt 
Andrea mentions multiple times that there is a default react component for something i made. How can we handle this for the demo and in the future? 

Can we get html/css from the react component somewhere and style where possible? Should we ignore the component completely and just build new html?

If you have time to walk by our office, that would be great :D

@afercia @JoseeWouters 
What i think i can do:
- Left menu details
- help links
- form fieldsets and legends
- form controlling labels
- Labels and legends should not contain extraneous content e.g. links
- general semantics (all back to divs)
- Descriptions


@IreneStr Is going to talk with the decision makers what we are going to do with accessibility for this PR, but also for our production process.

IreneStr:
@Sjardo I noticed you committed `packages/components/src/modal/components/.DS_Store`. Please remove it and add it to the .gitignore if needed 😉

Sjardo:
@afercia 

- I did some work on the checkboxes, radiobuttons, and text inputs.
- Fieldset and legends are applied to the checkboxes and radiobuttons
- Legends / labels do not consist of a help link. That's now next to it
- Description is moved to above the input field with an area ` aria-describedby"
- Removed html elements `section`, header`, `footer`, `article` and `nav`. I replaced them with plain `<divs>`
- external links have a `target="_blank"` on them
- Close button now has an `aria-label="Close"` on it

Could you do a check?

packages/components/src/modal/components/input/input.css
```diff
@@ -0,0 +1,30 @@
+:root {
+    --color-placeholder: #707070;
+}
+
+.yoast-field-group__textarea,
```
JoseeWouters: I'd like some global styling for textarea's and inputs too.

packages/components/src/modal/components/modal/modal.css
```diff
@@ -0,0 +1,297 @@
...
+
+.yoast-modal__heading-icon {
+    fill: var(--color-primary);
+    margin-right: 16px;
+    width: 19px;
+    height: 20px;
+    flex-shrink: 0;
+}
+
+.yoast-close {
```
JoseeWouters: Maybe we should move `yoast-close` to buttons.

packages/components/src/modal/components/radiobutton/radiobutton.css
```diff
@@ -0,0 +1,46 @@
+.yoast-field-group__radiobutton {
```
JoseeWouters: Same as for other input fields, I'd like some global styling where possible.

packages/components/src/modal/components/select/select.css
```diff
@@ -0,0 +1,23 @@
+.yoast-field-group__select {
```
JoseeWouters: Where possible, some global styling please

packages/components/src/modal/components/button/buttons.html
```diff
@@ -0,0 +1,13 @@
+<!--A button can both be used for a link and a button.
+Use a link when the button should redirect to another page
+use a button when it's used for an action: for example, opening a modal.-->
+
+<a href="${url}" class="yoast-button yoast-button--primary">${content}</a>
+
+<button class="yoast-button yoast-button--primary"></button>
```
afercia: We should clarify that buttons should never be empty:
1. preferred: add visible text within the button
2. alternatively: add screen-reader-text within the button
3. last resource: add an `aria-label="meaningful text"`

The cases 2 and 3 will still be problematic for speech recognition software users: they see an icon but they don't have a clue what the accessible name of the button is. Thus, they're not able to voice a proper command.

packages/components/src/modal/components/modal/modal.html
```diff
@@ -0,0 +1,95 @@
...
+	<link rel='stylesheet' href='../../components/input/input.css' type='text/css' media='all'/>
+	<link rel='stylesheet' href='../../components/data-model/data-model.css' type='text/css' media='all'/>
+	<link rel='stylesheet' href='../../components/button/buttons.css' type='text/css' media='all'/>
+	<link rel='stylesheet' href='../../components/modal/modal.css' type='text/css' media='all'/>
+
+	<link rel='stylesheet' href='../../base/utility.css' type='text/css' media='all'/>
+</head>
+<body class="yoast">
+<div class="yoast-modal__container">
+	<form class="yoast-modal" role="dialog">
```
afercia: 1. The `role="dialog"` should not be set on the `<form>` element as it overrides the native semantics of the form. Please add a wrapper element and set `role="dialog"` on it

There are other a11y requirements for modals, including labeling, focus management etc. Not sure if this should be part of the demo. Our React modal component already handles most (all?) of what's needed.

packages/components/src/modal/components/toggle/toggle.html
```diff
@@ -0,0 +1,54 @@
...
+
+	<link rel='stylesheet' href='../../components/field-group/field-group.css' type='text/css' media='all'/>
+	<link rel='stylesheet' href='../../components/toggle/toggle.css' type='text/css' media='all'/>
+
+</head>
+<body>
+<form>
+	<div class="yoast-field-group">
+
+		<label class="yoast-field-group__title">Label name</label>
```
afercia: Pending feedback on the usage of our existing React toggle component, which would be preferred.

PR #516, opened 2020-02-24
move social forms to separate package

### LINKED ISSUE #506: Move Social Preview forms to separate package
## Explanation and Technical Decisions
We want to move the SocialMetadataPreviewForms to their own package, so that they can be included in Free, separately from the Previews, which should be in Premium.

I think it is best to do this after https://github.com/Yoast/javascript/issues/505 is merged, so that you can also point to the new ReplacementVariableEditor package, but it is not a blocker per se.

## Tasks
- [ ] If it hasn't already done yet, merge `develop` into the feature branch to get the changes from #505 
- [ ] Move social metadatapreview forms and imageselect to a new package.
- [ ] If https://github.com/Yoast/javascript/issues/505 is already merged, point to the new ReplacementVariableEditor package.
- [ ] Notify Plugin 2 that a new package is coming to prevent surprises during monorepo releases.

## Feedback?
Discussed with @atimmer and @Xyfi 
@IreneStr

GuidoOffermans:
## Summary

* Created a new package `social-metadata-forms`.
* Initialized yarn, babel and eslint.
* Moved `SocialMetadataPreviewForm`, `ImageSelect`, redux (reducer, actions, selectors), and tests to this new package.
 

* Notify Plugin 2 that a new package is coming to prevent surprises during monorepo releases. (Do this when all is both packages are extracted.)

## Test instructions

* Check if the tests still work.
* Check the testApp if everything still works.

## Impact check

  * Social Appearance
  * Snippet preview

## UI changes

## Quality assurance

Fixes #506


packages/social-metadata-forms/package.json
```diff
@@ -0,0 +1,40 @@
...
+		"url": "https://github.com/Yoast/javascript.git",
+		"directory": "packages/social-metadata-forms"
+	},
+	"author": "Team Yoast",
+	"license": "GPL-3.0",
+	"private": false,
+	"scripts": {
+		"prepublishOnly": "rm -rf dist && cp -R src dist && cp package.json dist/package.json && json -I -f dist/package.json -e \"this.main='index.js'\" && cp .babelrc dist/.babelrc",
+		"test": "jest",
+		"lint": "eslint ."
```
bintzandt: I would prefer if we add a `--max-warnings=0` to this line. That makes sure that eslint will error when there would be any warnings.

This helps keep our code to the eslint standards since you would be forced to fix the warnings (or explain why you did not fix them).

By default, eslint does not error when there are any warnings in the files,

PR #518, opened 2020-02-25
Added webpack configuration to bundle css in @yoast/components

### LINKED ISSUE #517: Bundle css in @yoast/components
PRs: 
https://github.com/Yoast/wordpress-seo/pull/14548
https://github.com/Yoast/javascript/pull/549

Context: For the modal demo app and the admin redesign, team frontend needs a way to bundle the css from the monorepo and use it in the plugin. All css can be bundled in 1 file.

Also see https://github.com/Yoast/javascript/pull/518#issuecomment-596498253

## To do
- [x] Close https://github.com/Yoast/javascript/pull/518
- [x] Create a js file for each of the components for which HTML/CSS has been written.
    - [x] Import the corresponding CSS to those files.
- [x] Implement a CSS loader in the webpack config that bundles all CSS file into 1 CSS file.
  - [x] Use `css-loader` to load css into js files.
  - [x] Use [mini-css-extract](https://webpack.js.org/plugins/mini-css-extract-plugin/#extracting-all-css-in-a-single-file) to extract the css from the js files to one single file
- [x] Use the css file that has been bundled in the plugin.
- [x] Document the changes in the Readme.md of the monorepo.
     - Communicate to MaxServe that this has been changed
     - Communicate in #dev-plugin that this has been changed
- [ ] Ask @igorschoester to do a major release of `@yoast/components` after this PR has been merged (when it's time to de a release. There is no extra release needed).

Xyfi:
## Summary

* [@yoast/components] Adds css bundling.

* Styles are output to a `styles.css` file in the root of the `@yoast/components` package.

## Test instructions

* Add css files on any place in the `src` directory.
* Run `yarn build`.
* Check that all the styles are added to `styles.css` in the dist folder.

## Impact check

  * N/A

## UI changes

## Quality assurance

Fixes #517

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -14,8 +14,9 @@
   "scripts": {
     "test": "jest",
     "lint": "eslint . --max-warnings=96",
-    "build": "webpack && rm dist/main.js",
-    "prepublishOnly": "rm -rf dist && mkdir dist && find ./src/ -name \\*.js -exec rsync -R {} dist \\; && cp package.json dist/package.json && json -I -f dist/package.json -e \"this.main='src/index.js'\" && cp .babelrc dist/.babelrc && yarn build"
+    "build-css": "webpack && rm dist/main.js",
+    "watch-css": "webpack --watch",
+    "prepublishOnly": "rm -rf dist && mkdir dist && find ./src/ -name \\*.js -exec rsync -R {} dist \\; && cp package.json dist/package.json && json -I -f dist/package.json -e \"this.main='src/index.js'\" && cp .babelrc dist/.babelrc && yarn build-css && echo 'Finished'"
   },
   "jest": {
     "testRegex": ".*Test.js$",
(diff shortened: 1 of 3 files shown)


bintzandt:
Current status: a css file with all css of the package is generated. However, it cannot be used in development yet. You can watch the css file changes, but that built css needs to be copied to the plugin overr and over again, because PHP cannot resolve symlinks, so it cannot be used with the linked monorepo.

atimmer:
Alexander asked what the best way is to solve this problem. I will put my current understanding of the problem below.

**Problem**: We want to generate 1 CSS file for the plugin, so that file can be used for pages where there are no React components. We want users of React components to only receive the CSS that is actually required for those React components.

**Solution**: At the top of React components we put an `@import "./css/component.css`. Then we add the `css-loader` where the React components are used. This will result in 1 CSS file for the plugin, because the plugin imports the `index.js` and the `index.js` imports all components. And this will also work if you only import 1 component. Only downside is that this is a breaking change for all consumers of the packages. But that is essentially impossible to avoid when making such a breaking change to the CSS tooling. I would say all affected packages do a major release and mention the breaking change in the changelog.

PR #530, opened 2020-03-04
Change Dutch exception list checks to accommodate different type of matches

agnieszkaszuba:
## Summary

* [yoastseo] Changes exception list checks to take into account three different possible sub-lists: exact match lists, ending match lists, and verb lists

## Test instructions

* Find the following lists in the morphology data file: wordsNotToBeStemmedExceptions, stemmingExceptionStemsWithFullForms, stemExceptions.noVowelOrConsonantDoubling, stemTjeAndOnePrecedingVowel, wordsStemOnlyEnEnding
* Put examples of words that would be matched on these exception lists into the specs (determineStemSpec and/or specs for individual files where the lists are checked)
* If a word is on an ending match sub-list, put both exact matches and words ending in an entry from the list into specs. For example, if _kat_ is on the list, put both _kat_ and words like _meerkat_ or _binnenkat_ into specs.
* When checking a verb list, a match will be found either if the word is an exact match, or if the matched part of the word is preceded by a verb prefix from the compoundVerbsPrefixes object in the morphology data file. For example, if the verb _rennen_ is on the list, both _rennen_ and _aanrennen_ will be matched, but not _paardenrennen_ because _paarden_ is not a valid verb prefix.

## Quality assurance

Fixes #


manuelaugustin:
CR for documentation: I think it would be great to have some examples in the documentation, so that you get a better idea what kind of words a specific function is operating on.

manuelaugustin:
CR & acceptance 👍 

I tested with the following example words:


|             | stemmingExceptionStemsWithFullForms | stemExceptions.noVowelOrConsonantDoubling | stemTjeAndOnePrecedingVowel | wordsStemOnlyEnEnding |
|-------------|-------------------------------------|-------------------------------------------|-----------------------------|-----------------------|
| exactMatch  | Doctorandus                         | Impermeabel                               | Omaatje                     | -                     |
| endingMatch | Abacus                              | Improvisator                              | Ballerinaatje               | Fjorden               |
| verbs       | browse                              | Versleutel                                | -                           | Herleiden             |

Issue #550, opened 2020-03-13
Roadmap: Create reactified components


bintzandt:
## Explanation
Team Front-end has provided us with loads of CSS and HTML files that we can use to create React components.

This issue tracks the status of the reactification of all the different components.

## What needs to be done
In short, the following needs to be done for every component. **Note: there can be several components in a single HTML file.**
1. Create a `.js` file with the same name as the HTML file. Make sure to start with a capital letter since this is the convention that we are using in the `@yoast/components` package.
2. Copy the HTML from the `.html` to the `.js`. Decide for yourself: can I use a functional component or does this component need a state and should I therefore create a Class component (the React version that we are using does not support hooks...).
3. Render the component in the example app.
4. Write tests for the component. These can probably be inspired by the tests that are already written in `@yoast/components`.
4. Profit 🎉 

## Needed now
- [x] Field-group
	- Issue: https://github.com/Yoast/javascript/issues/558
	- PR: https://github.com/Yoast/javascript/pull/551
- [x] Inputs
	- Issue: https://github.com/Yoast/javascript/issues/558
	- PR: https://github.com/Yoast/javascript/pull/551
- [x] Radiobutton
	- Issue: https://github.com/Yoast/javascript/issues/559
	- PR: #562 
- [x] Select
	- Issue: https://github.com/Yoast/javascript/issues/560

## Needed later
- [x] Button
	- Issue: https://github.com/Yoast/javascript/issues/563
- [x] Checkbox
	- Issue: https://github.com/Yoast/javascript/issues/564
- [x] Data-model
	- Issue: https://github.com/Yoast/javascript/issues/565
- [x] Modal
	- Issue: https://github.com/Yoast/javascript/issues/566
- [x] Toggle
	- Issue: https://github.com/Yoast/javascript/issues/567

PR #551, opened 2020-03-13
Add FieldGroup and Input components

### LINKED ISSUE #558: Create TextInput and TextArea inputs
Parent #550

## Explanation

This issue relates to the input components.

## Tasks
- [x] Create a React component from the HTML
- [x] Write tests for the component
- [x] Add the new component to the `Reactification` tab in the example components app

bintzandt:
## Summary
This PR adds the new `Input` components to `@yoast/components`.

* [yoast-components] Create new React components based on the HTML that was already provided.

* A `FieldGroup` component has been created that can be used to wrap the `Input` fields with a label, description and a `HelpIcon`.
* The `description` of the `FieldGroup` renders conditionally based on the length of the description. If the length is 0, we assume that no description has been provided and we will therefore not render a `<p>`.
* I've created separate files for `TextInput` and `TextArea`. This seemed like the most logical decision at the moment. If it turns out that there is a better way to categorize the components than please go ahead :)

## Tasks
- [x] Add a `TextInput` component
- [x] Add a `TextArea` component
- [x] Write tests for the `TextInput` component
- [x] Write tests for the `TextArea` component

## Test instructions

_Test instructions will be written when the PR is finished._
* Check that the Example components app works as expected

## Quality assurance

Parent #550 

Fixes #558


bintzandt:
Two question that still remain:
1. Do we want to merge this into the feature branch? Or do we want to merge this into develop directly?
1. What needs to happen if we want to change the export of the old `Input` component to the new `TextInput` component?

packages/components/src/inputs/TextArea.js
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Renders a textarea for use in our HTML forms.
+ *
+ * @param {object} props The properties required for rendering this component.
+ *
+ * @returns {React.Component} A react component that can be used in our forms.
+ */
+const TextArea = ( props ) => {
+	const id = getId( props.id );
```
maartenleenders: 🤔 I was thinking...
Is it not preferable to have `const id = props.id || null;` (or even have that as the default prop)?

That way `id` will only be set if there is one set explicitly.

If you are worried about similar items, you can always set the `key`.

maartenleenders: Ah nvm you need the `id` for the labels, my bad

packages/components/src/inputs/TextInput.js
```diff
@@ -0,0 +1,94 @@
...
+			<input
+				id={ id }
+				name={ props.name }
+				defaultValue={ props.value }
+				type={ props.type }
+				className="yoast-field-group__inputfield"
+				aria-describedby={ props.ariaDescribedBy }
+				placeholder={ props.placeholder }
+				readOnly={ props.readOnly }
+				min={ props.min }
```
maartenleenders: I don't necessarily like passing irrelevant props to the input object, even if they are undefined (I believe they end up in the DOM as `<input id="bla" min max />`, correct?).

Isn't it better to not strictly set these props in advance, like we have now with the `undefined` defaultProps, but to have the input here saying:

```js
<input
	{ ...props }
/>
```
That way, if we don't add `min` or `max` to our `<TextInput id="bla" type="text" />` call, it won't appear in the DOM. Makes things a bit more readable, I think?

I might be missing something obvious, though?

maartenleenders:
CR: small suggestion, could you have a look and tell me what you think / why you wanted to do it differently?

Regarding your question, bit out of my depth but
1) social preview feature branch doesn't seem logical, we need text inputs elsewhere. A new feature branch also doesn't seem great.
If I am not mistaken, once our plugin can load the CSS, we can just export these new components without using them, so `develop` seems fine **if we don't release monorepo before our plugin is updated with the css-loader**. Let's discuss in the standup.
2) If they look exactly on the implementation side, perhaps not so much? I don't have an overview of how they differ currently.

packages/components/src/inputs/TextInput.js
bintzandt: The props **do** en up in the virtual DOM that React uses.

However, they are **not** present in the actual HTML DOM.

As to your suggestion: ESLint complains if there are props that are not required and that do not have a default value. Therefore, I have decided to provide a default value of `undefined` to make sure that they do not end up in the real (HTML) DOM.

If you like it better, I could opt for the following:
```jsx
<input
	id={ id }
	defaultValue={ props.value }
	className="yoast-field-group__inputfield"
	aria-describedby={ props.ariaDescribedBy }
	{ ...props }
/>
```

maartenleenders:
Acceptance 👍 

**BUT:** 
The `input.html` looks a bit different from the one in the example app. Not talking about the backgrounds so much, but about the borders. I think this is because the html loads the `borders.css` stylesheet (which loads some border styles into the root style), whereas we only import `input.css`. 

We should probably have a default css import helper somewhere, that will import all these related things. Do you agree?
 Holding the merge until we discuss 😉

maartenleenders:
Base styles issue is non-blocking, will be implemented later.

PR #562, opened 2020-03-19
Create RadioButtonGroup component

### LINKED ISSUE #559: Create RadioButton Component
Parent #550

## Explanation

This issue relates to the RadioButton component.

## Tasks
- [ ] Create a React component from the HTML
- [ ] Write tests for the component

bintzandt:
## Summary
This PR creates a `RadioButtonGroup` component based on the HTML provided by Team Front-end.

* [yoast-components] Adds a new `RadioButtonGroup` component.

* I have named the component `RadioButtonGroup` since it is used for rendering a group of radiobuttons (it does not make sense to render a single radiobutton imho).
* The options for the `RadioButtonGroup` can be passed via the `options` prop. I have decided to use the following object for the data: 
```js
{
	value: string,
	label: string,
	checked: bool,
}
```
A short explanation about the usage of the props:
- `value`: the value that will be submitted when this radiobutton is selected
- `label`: a label accompanying the radiobutton
- `checked`: an indicator whether this radiobutton is checked, optimally only one of the options should have this property set to `true`.
Furthermore, the `groupName` of every radiobutton can be passed to the `RadioButtonGroup` since it should be equal for every radiobutton in a group.

## Test instructions

1. Check that the implementation of the `RadioButtonGroup` in the example app works as expected.

## Quality assurance

Fixes #559

### DIFF
--- a/apps/components/ReactifiedComponentsWrapper.js
+++ b/apps/components/ReactifiedComponentsWrapper.js
@@ -43,8 +43,8 @@ const ReactifiedComponentsWrapper = () => {
 			<RadioButtonGroup
 				options={ [
 					{
-						value: "hoi",
-						label: "hoi",
+						value: "hey",
+						label: "hey",
 						checked: false,
 					},
 					{
@@ -54,7 +54,7 @@ const ReactifiedComponentsWrapper = () => {
 					}
 				] }
 				label="Horizontal radiobutton group"
-				groupName="groep2"
+				groupName="group1"
 			/>
 			<RadioButtonGroup
 				options={ [
@@ -71,7 +71,7 @@ const ReactifiedComponentsWrapper = () => {
 				] }
 				label="Vertical radiobutton group"
 				vertical={ true }
-				groupName="groep1"
+				groupName="group2"
 			/>
 		</div>
 	);
--- a/apps/components/ReactifiedComponentsWrapper.js
+++ b/apps/components/ReactifiedComponentsWrapper.js
@@ -45,28 +45,25 @@ const ReactifiedComponentsWrapper = () => {
 					{
 						value: "hey",
 						label: "hey",
-						checked: false,
 					},
 					{
 						value: "hi",
 						label: "hi",
-						checked: true,
 					}
 				] }
 				label="Horizontal radiobutton group"
 				groupName="group1"
+				selected="hi"
 			/>
 			<RadioButtonGroup
 				options={ [
 					{
 						value: "haha",
 						label: "haha",
-						checked: false,
 					},
 					{
 						value: "hoho",
 						label: "hoho",
-						checked: true,
 					}
 				] }
 				label="Vertical radiobutton group"
--- a/apps/components/ReactifiedComponentsWrapper.js
+++ b/apps/components/ReactifiedComponentsWrapper.js
@@ -1,6 +1,7 @@
 import React from "react";
 import TextInput from "@yoast/components/src/inputs/TextInput";
 import TextArea from "@yoast/components/src/inputs/TextArea";
+import RadioButtonGroup from "@yoast/components/src/radiobutton/RadioButtonGroup";
 
 const ReactifiedComponentsWrapper = () => {
 	return (
@@ -38,6 +39,41 @@ const ReactifiedComponentsWrapper = () => {
 				placeholder="Ugly placeholder"
 				value="Wow, what happens now??"
 				description="The greatest textarea ever!!1!"
+			/>
+			<RadioButtonGroup
+				options={ [
+					{
+						groupName: "groep2",
+						value: "hoi",
+						label: "hoi",
+						checked: false,
+					},
+					{
+						groupName: "groep2",
+						value: "hi",
+						label: "hi",
+						checked: true,
+					}
+				] }
+				label="Horizontal radiobutton group"
+			/>
+			<RadioButtonGroup
+				options={ [
+					{
+						groupName: "groep1",
+						value: "haha",
+						label: "haha",
+						checked: false,
+					},
+					{
+						groupName: "groep1",
+						value: "hoho",
+						label: "hoho",
+						checked: true,
+					}
+				] }
+				label="Vertical radiobutton group"
+				vertical={ true }
 			/>
 		</div>
 	);
--- a/apps/components/ReactifiedComponentsWrapper.js
+++ b/apps/components/ReactifiedComponentsWrapper.js
@@ -41,14 +41,15 @@ const ReactifiedComponentsWrapper = () => {
 				description="The greatest textarea ever!!1!"
 			/>
 			<RadioButtonGroup
+				onChange={ console.log }
 				options={ [
 					{
-						value: "hey",
-						label: "hey",
+						value: "hey-value",
+						label: "Hey there!",
 					},
 					{
-						value: "hi",
-						label: "hi",
+						value: "hi-value",
+						label: "Hi there!",
 					}
 				] }
 				label="Horizontal radiobutton group"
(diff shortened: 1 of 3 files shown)


packages/components/src/radiobutton/RadioButtonGroup.js
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Function that returns a radiobutton with accompanying label.
+	 *
+	 * @param {string} value The value of the radiobutton.
+	 * @param {string} label The label that accompanies the radiobutton.
+	 * @param {boolean} checked Whether the radiobutton is checked.
+	 *
+	 * @returns {React.Component} A single radiobutton field with label.
+	 */
+	const RadioButtonWithLabel = ( { value, label, checked } ) => <Fragment>
```
Xyfi: I would prefer `LabeledRadioButton` since it prevents the use of `With` in the name of the component.

Xyfi: Also, please define this component outside of `RadioButtonGroup`, because it will have to redefine this component each time `RadioButtonGroup` has to be rerendered.

packages/components/src/radiobutton/RadioButtonGroup.js
```diff
@@ -0,0 +1,108 @@
...
+			onChange={ props.onChange }
+		/>
+		<label
+			htmlFor={ value }
+		>
+			{ label }
+		</label>
+	</Fragment>;
+
+	RadioButtonWithLabel.propTypes = {
```
Xyfi: You can put this definition in a variable so you can reuse it in the `options` PropType definition of `RadioButtonGroup`. You could name it `radioButtonProps` for example.

packages/components/src/radiobutton/RadioButtonGroup.js
```diff
@@ -0,0 +1,108 @@
...
+		label: PropTypes.string.isRequired,
+		checked: PropTypes.bool.isRequired,
+	};
+
+	/**
+	 * Returns a set of radiobuttons that are on the same line.
+	 *
+	 * @returns {React.Component} A div with radiobuttons in it.
+	 */
+	const HorizontalRadioButtons = () =>
```
Xyfi: Also, please define this component outside of `RadioButtonGroup`. See above.

packages/components/src/radiobutton/RadioButtonGroup.js
```diff
@@ -0,0 +1,108 @@
...
+		<div className="yoast-field-group__radiobutton">
+			{ props.options.map( option => <RadioButtonWithLabel key={ option.value }{ ...option } /> ) }
+		</div>;
+
+	/**
+	 * Returns a set of radiobuttons that are vertically aligned.
+	 *
+	 * @returns {React.Component} A lot of divs with one radiobutton in it.
+	 */
+	const VerticalRadioButtons = () =>
```
Xyfi: Also, please define this component outside of `RadioButtonGroup`. See above.

packages/components/src/radiobutton/RadioButtonGroup.js
Xyfi: Also I wonder whether creating a separate `HorizontalRadioButtons` and `VerticalRadioButtons` is necessary since the only difference is a className, so you could choose to optionally add the className based on the `vertical` prop.

packages/components/src/radiobutton/RadioButtonGroup.js
```diff
@@ -0,0 +1,108 @@
...
+				className="yoast-field-group__radiobutton yoast-field-group__radiobutton--vertical"
+				key={ option.value }
+			>
+				<RadioButtonWithLabel { ...option } />
+			</div>,
+			) }
+		</Fragment>;
+
+	const id = getId( props.id );
+	const fieldGroupProps = {
```
Xyfi: You pass all props down to `FieldGroup` even though this is not necessary for most props. I would extract all props that are not needed to be passed down using:
```
const {
  prop1,
  prop2,
  ...fieldProps
} = props;
```

packages/components/src/radiobutton/RadioButtonGroup.js
```diff
@@ -0,0 +1,108 @@
...
+	return (
+		<FieldGroup
+			{ ...fieldGroupProps }
+		>
+			{ props.vertical ? <VerticalRadioButtons /> : <HorizontalRadioButtons /> }
+		</FieldGroup>
+	);
+};
+
+RadioButtonGroup.propTypes = {
```
Xyfi: Since only one value can be selected, I would opt to have a `selected` prop, instead of manually setting a `checked` props for each radioButton. This makes managing the selected radiobutton a lot easier.

apps/components/ReactifiedComponentsWrapper.js
```diff
@@ -39,6 +40,39 @@ const ReactifiedComponentsWrapper = () => {
 				value="Wow, what happens now??"
 				description="The greatest textarea ever!!1!"
 			/>
+			<RadioButtonGroup
```
Xyfi: Please keep variables and string used in tests in English. Like `groep1` and `hoi`.

packages/components/src/radiobutton/RadioButtonGroup.js
bintzandt: It is not only the className that differs. In one case, we have the following structure:
```jsx
<div className="yoast-field-group__radiobutton">
	<LabeledRadioButton />
	<LabeledRadioButton />
	<LabeledRadioButton />
</div>
```
In the other we have:
```jsx
<Fragment>
	<div className="yoast-field-group__radiobutton yoast-field-group__radiobutton--vertical">
		<LabeledRadioButton />
	</div>
	<div className="yoast-field-group__radiobutton yoast-field-group__radiobutton--vertical">
		<LabeledRadioButton />
	</div>
	<div className="yoast-field-group__radiobutton yoast-field-group__radiobutton--vertical">
		<LabeledRadioButton />
	</div>
</Fragment>
```

packages/components/src/radiobutton/RadioButtonGroup.js
bintzandt: I'm not sure how you want to implement this. Simply changing the prop `checked` to `selected` will not make much of a difference. Or do you mean that I should provide the prop `selected` to the `RadioButtonGroup`?

If so, what value should it get? The index of the selected radiobutton in the array? Or the id of the selected radiobutton?

packages/components/src/radiobutton/RadioButtonGroup.js
bintzandt: I don't think the destructuring of `fieldProps` in the assignment works as expected.

packages/components/src/radiobutton/RadioButtonGroup.js
bintzandt: If I place this component outside of the RadioButtonGroup, I have to provide additional props to it. This means that I can no longer apply your suggestion.
> You can put this definition in a variable so you can reuse it in the `options` PropType definition of `RadioButtonGroup`. You could name it `radioButtonProps` for example.

At least, not that I can see.

bintzandt:
Applied a small part of the CR but I have some additional questions. If @Xyfi has time tomorrow, it might be easier to discuss this "in person" via Zoom or a Slack call 🙂

packages/components/src/radiobutton/RadioButtonGroup.js
maartenleenders: Seems like we could ask team frontend to rearrange the css slightly, so that you only need ```html
	<div className="yoast-field-group__radiobutton yoast-field-group__radiobutton--vertical">
		<LabeledRadioButton />
		<LabeledRadioButton />
		<LabeledRadioButton />
	</div>
```

Right?

bintzandt: > Seems like we could ask team frontend to rearrange the css slightly, so that you only need
> 
> ```
> <div className="yoast-field-group__radiobutton yoast-field-group__radiobutton--vertical">
> 	<LabeledRadioButton />
> 	<LabeledRadioButton />
> 	<LabeledRadioButton />
> </div>
> ```
> 
> Right?

Yes, that might be. But lets not bother front-end too much at this moment. I believe that Josee has enough work as is.

Xyfi:
CR 👍 
Acceptance 🚧 

In the components app I get this error:

Screenshot 2020-03-25 at 13 20 43

bintzandt:
Should be fixed in the latest commit.

Please verify and it should be good to go 🚀

PR #569, opened 2020-03-20
Add CheckboxGroup component

### LINKED ISSUE #564: Create Checkbox component
Parent #550

## Explanation

This issue relates to the Checkbox component.

## Tasks
- [x] Create React components from the HTML
- [x] Write tests for the components
- [x] Add the new component to the `Reactification` tab in the example components app

bintzandt:
## Summary
Converts the HTML from Team Front-end to a working React component that generates a group of checkboxes with the correct classes.

* [yoast-components] Add a new CheckboxGroup component.

## Test instructions

* Check that the checkboxes in the example components app work and look as expected.

## Quality assurance

Fixes #564

### DIFF
--- a/packages/components/src/checkbox/CheckboxGroup.js
+++ b/packages/components/src/checkbox/CheckboxGroup.js
@@ -1,0 +1,112 @@
+import React, { Fragment } from "react";
+import PropTypes from "prop-types";
+import FieldGroup, { FieldGroupDefaultProps, FieldGroupProps } from "../field-group/FieldGroup";
+import { getId } from "../GenerateId";
+// Import required CSS.
+import "./checkbox.css";
+
+/**
+ * Smallest component in the CheckboxGroup. Contains a label and the checkbox.
+ *
+ * @param {string} id The ID of te checkbox.
+ * @param {string} label The label of the checkbox.
+ *
+ * @returns {React.Component} A React component that wraps around the HTML checkbox.
+ */
+const Checkbox = ( { id, label } ) => <Fragment>
+	<input type="checkbox" id={ id } />
+	<label htmlFor={ id } className="yoast-field-group__checkbox">{ label }</label>
+</Fragment>;
+
+Checkbox.propTypes = {
+	label: PropTypes.string.isRequired,
+	id: PropTypes.string.isRequired,
+};
+
+/**
+ * Component that renders a vertical list of checkboxes with their labels.
+ *
+ * @param {Object[]} options An array of options of the checkboxes.
+ *
+ * @returns {*} A React component that contains a list of vertical checkboxes.
+ */
+const VerticalCheckboxes = ( { options } ) => {
+	return options.map( option => {
+		const id = getId( option.id );
+		return (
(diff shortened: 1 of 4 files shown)


packages/components/src/checkbox/CheckboxGroup.js
```diff
@@ -0,0 +1,138 @@
...
+		const id = getId( option.id );
+		return (
+			<div key={ id } className="yoast-field-group__checkbox">
+				<Checkbox id={ id } label={ option.label } onChange={ onChange } { ...option } />
+			</div>
+		);
+	} );
+};
+
+VerticalCheckboxes.propTypes = {
```
Xyfi: The vertical and horizontal checkbox prop-types are the same. Put them in a separate variable.

packages/components/src/checkbox/CheckboxGroup.js
```diff
@@ -0,0 +1,138 @@
...
+	options: PropTypes.arrayOf( PropTypes.shape( {
+		label: PropTypes.string.isRequired,
+		id: PropTypes.string,
```
Xyfi: I'd say id should be required, because otherwise there is no way of identifying the selected value.

packages/components/src/checkbox/CheckboxGroup.js
```diff
@@ -0,0 +1,138 @@
...
+			{ vertical
+				? <VerticalCheckboxes { ...checkboxProps } />
+				: <HorizontalCheckboxes { ...checkboxProps } />
+			}
+		</FieldGroup>
+	);
+};
+
+
+CheckboxGroup.propTypes = {
```
Xyfi: Instead of passing `selected` for each option individually, I would opt to pass a `selected` array of strings that correspond to the selected checkboxes' ids.

This on change should also return an array of selected values.

As a rule of thumb: A onChange function should always return the same value as the prop that is used to identify what the entered value is.

packages/components/src/checkbox/CheckboxGroup.js
```diff
@@ -0,0 +1,138 @@
...
+	} = props;
+	const componentId = getId( id );
+
+	const checkboxProps = {
+		options,
+		onChange: event => onChange( event.target.id ),
+	};
+
+	return (
+		<FieldGroup
```
Xyfi: The onChange value can be set on this component, instead of the underlying individual components.

bintzandt: Not true. Since `FieldGroup` is a custom component, it does not do anything if we set the `onChange` prop there. Unless we change how the `FieldGroup` component work.

I added an additional `div` to which I assigned the `onChange` event.

GuidoOffermans:
### CR. 
Code looks good!

### Acceptance

Works perfectly in the components-app. 

But looks a bit funky in the plugin, may need tweaking.
Screenshot 2020-04-01 at 10 21 42

maartenleenders:
Styling issues are picked up by @JoseeWouters

PR #582, opened 2020-03-30
Avoids verbal stemming of Spanish non-verbs that look like verbs

manuelaugustin:
## Summary

* [yoastseo] Avoids verbal stemming of Spanish non-verbs that look like verbs.

* I've added the exception check before the stemming for verbs. I didn't check, however, if the check actually applies in all cases. For example, we have a lot of words ending in `idad` in the exception list, but the stemmer stems `idad` as a derivational noun suffix before we get to the exception check. For this, we could have clean-up later, so that we don't have unnecessarily long exception lists. That's only a performance consideration, however. It should not negatively affect the accuracy of the stemmer.
* For specs I left all forms commented out that don't get correctly stemmed. This affects a number of plurals of forms included in the lists. 

## Test instructions

- Run `yarn test` and check that all tests pass.

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LIN-310

### DIFF
--- a/packages/yoastseo/spec/morphology/spanish/stemSpec.js
+++ b/packages/yoastseo/spec/morphology/spanish/stemSpec.js
@@ -174,6 +174,146 @@ const wordsToStem = [
 	[ "desconsentir", "desconsent" ],
 	[ "desconsiento", "desconsent" ],
 	[ "desconsintió", "desconsent" ],
+	// Words that look like verb forms but aren't verbs.
+	// Non-verb ending in -ió
+	[ "chevió", "chevi" ],
+	// [ "cheviós", "chevi" ],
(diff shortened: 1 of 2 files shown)


nataliashitova:
CR 👍 I've only implemented a tiny fix to improve code readability.
I created an issue to add the results of the remaining two research issues to the list https://yoast.atlassian.net/browse/LIN-342.

I've also noticed that not all specs pass, which represent plural forms of the words added to the exception lists. We need to add plural forms for some forms from the list, which are not yet added to the list `nonPluralsOnS`. I've created an issue for that https://yoast.atlassian.net/browse/LIN-341.

PR #584, opened 2020-03-30
Create Select components

### LINKED ISSUE #560: Create Select component
Parent #550

## Technical decisions
We are going to use the `select2` package in all the places where we are currently using it. This means that we will have to create 2/3 React components:
- A select for the normal use case
- A select for `multiple="multiple"`
- A select for when the `select2` package is used

Depending on the implementation, the first two can be taken together in a single React component.

## Explanation

This issue relates to the Select component.

## Tasks
- [ ] Create a React component from the HTML
- [ ] Write tests for the component

bintzandt:
## Summary
This PR adds two new components: `Select` and `MultiSelect`. `Select` is a wrapper around the default HTML select while `MultiSelect` uses the `select2` package to create a select.

* [yoast-components] Adds a `Select` and `MultiSelect` component.

* I have tried to create an abstraction for overlap of code between the `MultiSelect` and the `Select` components. However, I have been unable to create a suitable one without any errors. If someone has a better idea, feel free to try it.

## Test instructions

* 🔥-up the example components app and checkout the `Select` and `MultiSelect`. Check that they behave as expected.

## Quality assurance

Fixes #560


bintzandt:
## In WordPress
I have tested these components in WordPress, these are my conclusions:
- The styling of the `Select` and `MultiSelect` components in the plugin is off
	- This happens with and without [@JoseeWouters's new branch](https://github.com/Yoast/javascript/pull/593).
- When testing with [the new branch](https://github.com/Yoast/javascript/pull/593), the styling of other components changes as well.

packages/components/src/select/Select.js
```diff
@@ -0,0 +1,185 @@
...
+	}
+}
+
+MultiSelect.propTypes = selectProps;
+MultiSelect.defaultProps = selectDefaultProps;
+
+/**
+ * React wrapper for a basic HTML select.
+ */
+export class Select extends React.Component {
```
maartenleenders, line 122: Is there a reason we are not exporting this in the components index yet?

maartenleenders:
Looks good to me, though styling wise the selects do not take the full width of the containing div. Not sure if that's ok, but the styling will be looked at anyways by @JoseeWouters .

Could you update the README to reflect the need to include `jQuery` and `select2` scripts in the header?

apps/components/index.html
```diff
@@ -5,6 +5,14 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Yoast Components</title>
     <link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/draft-js/0.10.5/Draft.min.css">
+    
+    <!-- Scripts and styles required for displaying the selects in the Reactified Componentens tab -->
```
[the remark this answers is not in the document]
bintzandt: Yes, it does. I'll add that to our documentation.

packages/components/src/select/Select.js
bintzandt: Yes, this gives some feedback that these components should not be used yet because they still might change. Once everything is ready, the idea is to remove the old components (where applicable) and export the new components from the index of the package.

bintzandt:
> Looks good to me, though styling wise the selects do not take the full width of the containing div. Not sure if that's ok, but the styling will be looked at anyways by @JoseeWouters .
> 
> Could you update the README to reflect the need to include `jQuery` and `select2` scripts in the header?

Hmm, might have done a little too much.

Anyway, the `MultiSelect` now checks whether `jQuery` and `select2` are available. If one of them is missing, it throws an `ErrorWithUrl` in the constructor. This error is then caught by the `ErrorBoundary` that displays an error message with a help URL instead.

To make sure that we do not catch to many errors in the error boundary, it only catches errors that have type set to `YoastError`. This is exactly the case for the `ErrorWithUrl` since it extends the `YoastError` class which sets that type.

Welp...

apps/components/ReactifiedComponentsWrapper.js
```diff
@@ -107,6 +108,34 @@ const ReactifiedComponentsWrapper = () => {
...
+				name="my-selection"
+				options={
+					[
+						{ name: "option 1", value: "opt1" },
+						{ name: "option 2", value: "opt2" },
+						{ name: "option 3", value: "opt3" },
+						{ name: "option 4", value: "opt4" },
+					]
+				}
+				selected={ "opt1" }
```
GuidoOffermans, line 123: It was a bit confusing to me that the selected prop was not an array. Later by reading te code it makes sense though.

apps/components/ReactifiedComponentsWrapper.js
```diff
@@ -107,6 +108,34 @@ const ReactifiedComponentsWrapper = () => {
...
+				name="my-selection"
+				options={
+					[
+						{ name: "option 1", value: "opt1" },
+						{ name: "option 2", value: "opt2" },
+						{ name: "option 3", value: "opt3" },
+						{ name: "option 4", value: "opt4" },
+					]
+				}
+				selected={ [ "opt1" ] }
```
GuidoOffermans, line 137: Same here. With a select i guessed it need just one option. The array confused me. But if we use the same props for both, do a strict array input?

GuidoOffermans:
### CR 
Some minor ideas about the input for `selected` props.

GuidoOffermans:
### ACC
Works great in the component app! 🎉

GuidoOffermans:
### ACC in plugin
It works like in the components app. The styling just looks off.
Screenshot 2020-04-06 at 13 39 40

PR #606, opened 2020-04-08
Add an external stemmer for Russian

nataliashitova:
## Summary

* [yoastseo] Adds an external stemmer for the Russian language.

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LIN-364


manuelaugustin:
CR & acceptance 👍

I noticed that the check for the derivational ending was removed compared to the base stemmer. Since that was only one ending, I suppose it was not a very valuable part of the stemmer and caused inaccuracies?

PR #616, opened 2020-04-09
Improve the way keyphrase distribution assessment works with lists

manuelaugustin:
## Summary

* [yoastseo] Makes sure that lists with single words don't skew the keyphrase distribution score.
* [Yoast SEO Premium] Fixes a bug when the keyphrase distribution assessment might return skewed results if a text contains lists of single words.

* I've chosen a very straightforward solution, simply removing all list tags from the text before it gets processed further for keyphrase distribution.
* One drawback of the new functionality is that list items that are sentences and don't end in a punctuation mark now are merged with the following list item. In longer lists, this can lead to very long sentences, which means that false positives might be likely. However, this will lead to a keyphrase distribution result that is too positive in some cases. While not a completely accurate result, it will not cause as frustration for the user as an undeserved negative result.

## Test instructions

* Have a look at the specs. Think about whether you can come up with any scenarios not yet covered.
* Add a relatively short text to your paper (but long enough to trigger keyphrase distribution). Add a keyphrase and instances of that keyphrase in the text. Add a list with single words as list items. Make sure that your keyphrase distribution score doesn't get worse after adding the list of single words.

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LIN-333


omarreiss:
@manuelaugustin I'd like to see some more unit tests for this. Could you also cover:
- cases where every listitem is essentially a paragraph. You could take recipe instructions as an example.
- cases where paragraph length listitems are mixed with sentence length listitems. Real world example could be an action list underneath meeting notes. Some actions are short sentences. Some actions are more complex and consist out of a few steps / sentences.

I think if we cover this in unit tests we will see issues faster.

manuelaugustin:
@omarreiss I've added the examples you suggested (lists consisting of paragraphs & mixed lists consisting of both paragraphs and sentences), see my latest commit: https://github.com/Yoast/javascript/pull/616/commits/acfa1ee1792774b9570ec2b93140ed6dfc6f5fac

Let me know if there are any more examples I should add, or whether I should proceed to the implementation.

omarreiss:
I would like to see testcases that are a little more real world. If I look at a random recipe online for grilled cauliflower, i get the following instruction list:
```
1. Remove the outer leaves from each cauliflower head. Cut off the bottom stem end so that you create a flat base and can stand the cauliflower up on a cutting board. Resting the cauliflower on the stem, use a large, sharp knife to trim away the sides, then cut the remaining head into 2 very thick or 3 more moderate "steaks." Reserve the sides for another use (see notes). In a small bowl, stir together the olive oil, lemon zest and lemon juice, garlic, and honey.
2. Heat a grill to medium (about 350 degrees F). Brush one side of each cauliflower steak with the lemon–olive oil mixture and then sprinkle the brush sides with half of the salt. Place the salted side down on the hot grill, then brush the tops of the steaks with the olive oil mixture and season with the remaining salt. Cover the grill and let cook for 5 to 6 minutes, until the bottom is beginning to char. Flip the cauliflower, then re-cover the grill and cook 5 additional minutes, until the cauliflower is tender. Remove from the grill. Sprinkle with the red pepper flakes, parsley, and walnuts. Serve hot with lemon wedges.
```

Just an example, but these are much longer and more realistic examples than what is in the unit tests now. I would prefer you take real world examples like this to have representative tests. I guess it's clear what I mean now, so feel free to improve your unit tests based on this feedback and move straight over to implementation.

agnieszkaszuba:
CR & acceptance: 👍 
Everything works and I cannot think of any other scenarios that the current specs don't cover :)

PR #622, opened 2020-04-15
Adding stemmer for Portuguese

hannaw93:
## Summary

* [yoastseo] adds a stemmer for Portuguese

## Test instructions

* Make sure all the specs are passing
* Make sure the coverage test returns only 10 errors and that they concern words ending in -_ção_ and -_ções_

## Relevant technical choices

* This stemmer is based on [this](https://github.com/NaturalNode/natural/blob/master/lib/natural/stemmers/porter_stemmer_pt.js) implementation of the Porter algorithm, but with code style adapted to resemble the [Spanish Porter stemmer](https://github.com/NaturalNode/natural/blob/master/lib/natural/stemmers/porter_stemmer_es.js)
* The current coverage test is based on the word and stem pairs available on the [snowball website](https://snowballstem.org/algorithms/portuguese/stemmer.html). 
* There are 10 stems that are different in the list of stems from the snowball website compared with the stems the current stemmer produces. These are words with the suffixes -_uço~es_ and -_uça~o_. The stemming of those suffixes was commented out in the source javascript stemmer. I try adding this piece of code and it fixes those 10 stems but creates problems with other words with those suffixes. So there seems to be a general problem with those suffixes, which we can look into further at a later point. 

## Quality assurance

Fixes https://yoast.atlassian.net/jira/software/projects/LIN/boards/4?selectedIssue=LIN-372


manuelaugustin:
CR & acceptance 👍

Issue #626, opened 2020-04-16
Upgrade outdated dependencies


bintzandt:
## Explanation

In the monorepo, there are some dependencies that are outdated. We should upgrade them in the future to ensure that bugs are fixed and we can profit from new functionality. 

There are also some packages that will need to be upgraded in order to have node v12 support (which is the current LTS release).

The list of outdated packages can be found using `yarn outdated`.

## Technical decisions

* `babel-cli`, `webpack-dev-server` and `jest` are the ones that *need* to be upgraded before we can have proper node v12 support.
* Remove the `resolutions` from `package.json` when the upgrade is complete since it is no longer needed.

PR #627, opened 2020-04-16
Add resolutions to package.json

bintzandt:
## Summary
This PR adds a resolutions section to the `package.json` in the root of the monorepo.

This `resolution` forces `fsevents` to be version `1.2.9`, which is the first version of `fsevents` that adds support for node v12.

You can find more information about yarn's `resolutions` [here](https://classic.yarnpkg.com/en/docs/package-json#toc-resolutions).

* [monorepo-root] Force `fsevents` to version `1.2.9` to add node 12 support.

* This is a work-around for the real problem, which is that our dependencies are heavily out-dated in some of our packages. Although this PR will allow us to install and build on node v12, we should upgrade our packages to add proper support.

  You can check the release cycle for node [here](https://nodejs.org/en/about/releases/). Node 10 will receive updates until 2020-05-19, after which the maintenance period starts (meaning that serious bugs / security flaws will still be fixed). EOL (End-Of-Life) for Node 10 is on 2021-04-30.
* The tests in `yoastseo` will fail when using node v12. This is currently also happening in Travis in a test that is allowed to fail.
* I have created [an issue](https://github.com/Yoast/javascript/issues/626) for upgrading the dependencies.

## Test instructions

* Check-out the develop branch.
* Switch to node v12 (I use `nvm` to do this easily)
* Run `yarn install`
* Check that it errors on `fsevents`, although the install will be successful since `fsevents` is optional.
* Switch to this branch
* Run `yarn install` (if it says "nothing changed" use `yarn install --force`)
* No errors 🎉 

## UI changes

## Quality assurance

Fixes N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,5 +34,8 @@
     "lerna-publish": "lerna publish --contents dist --yes from-git",
     "lerna-publish-rc": "lerna publish --contents dist --dist-tag rc --yes from-git",
     "list-deps": "node ./scripts/list-dependencies.js"
+  },
+  "resolutions": {
+    "**/fsevents": "1.2.9"
   }
 }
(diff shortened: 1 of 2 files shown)


bintzandt:
Closing this PR because Travis cannot cope with resolutions for Optional Dependencies.

PR #634, opened 2020-04-20
Rewire siteName prop to siteUrl and small refactors

maartenleenders:
## Summary

All context and test instructions can be found here: https://github.com/Yoast/wordpress-seo-premium/pull/2744

* [non-user-facing] Switch `siteName` to `siteUrl` to better reflect what the previews should output.
* [non-user-facing] Renamed `errors` field to `warnings` in redux state, because it is used to store warnings about the image.

These are non-user-facing because the packages have not been released yet.

Fixes https://github.com/Yoast/wordpress-seo-premium/issues/2734


Xyfi:
CR + Acceptance 👍

PR #637, opened 2020-04-21
Select text overflow fix

rolf-yoast:
## Summary

* Small fix for select element where the selected option text was too long for the container.

*

## Test instructions

* Pull "14544-multisite-dashboard-updates" from the "wordpress-seo" and the "javascript" repository.
* Link the monorepo.
* In the multisite environment click on "SEO".
* Check if the selected option doesn't show behind the arrow of the select component.

## Impact check

  *

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/packages/components/src/select/select.css
+++ b/packages/components/src/select/select.css
@@ -29,6 +29,7 @@
 .yoast-field-group select,
 .yoast-field-group__select select {
     line-height: 1.9;
+    padding-right: 40px;
 }
 
 .select2-container.select2-container--default {


packages/components/src/select/select.css
```diff
@@ -29,6 +29,7 @@
 .yoast-field-group select,
 .yoast-field-group__select select {
     line-height: 1.9;
+    padding-right: 40px;
```
JoseeWouters, line 32: Is this the best way to solve this? Adding a padding here could mean style changes at other places. If the component group needs padding, I'd like to solve this at a bit higher level.

rolf-yoast: This only changes the select element from:
image
to:
image
So it doesn't effect the component group but just the padding inside the select element so that the text won't show behind the arrow.

afercia:
CR&A 👍

PR #646, opened 2020-04-24
Move Polish stemming data to morphology data file

manuelaugustin:
## Summary

* [yoastseo] Moves the Polish stemming data to a newly created Polish morphology data file.

* I also enabled support in `getLanguagesWithWordFormSupport` because that's needed to test whether the morphology data file gets properly picked up in the content analysis app.
* Enabling morphology support meant that the behavior of some unit tests that previously assumed no morphology support for Polish changed. Those were adapted to reflect the new situation.

## Test instructions

* Check the unit tests
  * Run `yarn test` and make sure that all tests still pass.
  * Make sure that the coverage is still the same as before the change.
* Test in the content analysis app
  * Start the content analysis app by going to `javascript/apps/content-analysis` and run `yarn start`.
  * In the content analysis app, make sure that the locale is set to the right language and that the toggle for morphology support is switched on.
  * Add a text of at least 150 words.
  * Set a keyphrase and use various (supported) forms of that keyphrase in the text.
  * Make sure that all forms get recognized in the keyphrase density assessment (click on the eye marker next to the result and confirm that you can see all forms marked below).

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LIN-395

### DIFF
--- a/packages/yoastseo/src/morphology/polish/stem.js
+++ b/packages/yoastseo/src/morphology/polish/stem.js
@@ -20,167 +20,6 @@
  * SOFTWARE.
  */
 
-const diminutiveSuffixes = {
-	suffixes1: {
-		wordEndings: [ "eczek", "eczka", "eczki", "iczek", "iczka", "iczki", "iszek", "iszka", "iszki", "aszek", "aszki",
-			"aszka", "uszek", "uszka", "uszki", "aczek", "aczka", "aczki" ],
-		wordShouldBeLongerThan: 6,
-		suffixLength: 5,
-	},
-	suffixes2: {
-		wordEndings: [ "ątko", "unio", "usia" ],
-		wordShouldBeLongerThan: 6,
-		suffixLength: 4,
-	},
-	suffixes3: {
-		wordEndings: [ "enek", "enka", "enki", "enką", "ejek", "ejka", "ejki", "ejką", "erek", "erki" ],
-		wordShouldBeLongerThan: 6,
-		suffixLength: 2,
-	},
-	suffixes4: {
-		wordEndings: [ "ek", "ak", "ka", "ko", "uś" ],
-		wordShouldBeLongerThan: 4,
-		suffixLength: 2,
-	},
-};
-
-const verbSuffixes = {
-	suffixes1: {
-		wordEndings: [ "łybyście", "libyście" ],
-		wordShouldBeLongerThan: 10,
-		suffixLength: 8,
-	},
-	suffixes2: {
-		wordEndings: [ "łybyśmy", "libyśmy" ],
-		wordShouldBeLongerThan: 9,
-		suffixLength: 7,
-	},
-	suffixes3: {
-		wordEndings: [ "liście", "łyśmy", "liśmy" ],
-		wordShouldBeLongerThan: 8,
-		suffixLength: 6,
-	},
-	suffixes4: {
-		wordEndings: [ "łabym", "łabyś" ],
-		wordShouldBeLongerThan: 7,
-		suffixLength: 5,
-	},
-	suffixes5: {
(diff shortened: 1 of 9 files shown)


packages/yoastseo/src/morphology/polish/stem.js
```diff
@@ -295,18 +138,18 @@ export default function stem( word ) {
...
 	}
 
 	if ( ! stemmedWord ) {
-		stemmedWord = findSuffixInClassAndStem( word, verbSuffixes );
+		stemmedWord = findSuffixInClassAndStem( word, morphologyData.externalStemmer.verbSuffixes );
 	}
 
 	if ( ! stemmedWord ) {
-		stemmedWord = stemAdjectivesAndAdverbs( word );
+		stemmedWord = stemAdjectivesAndAdverbs( word, morphologyData );
```
iolandasequino, line 152: As Manuel said, the path for this function is probably in the morphologyData file, but it would be more consistent to have it here

packages/yoastseo/src/morphology/polish/stem.js
```diff
@@ -262,15 +101,17 @@ const findSuffixInClassAndStem = function( word, suffixClass ) {
  * Stems adjective and adverb suffixes. After stemming the suffixes, looks for the superlative prefix 'naj' and stems it
```
iolandasequino, line 101: Do we want to have these example line in the stem function file?

packages/yoastseo/src/morphology/polish/stem.js
```diff
@@ -262,15 +101,17 @@ const findSuffixInClassAndStem = function( word, suffixClass ) {
...
  */
-const stemAdjectivesAndAdverbs = function( word ) {
-	const stemmedWord = findSuffixInClassAndStem( word, adjectiveAndAdverbSuffixes );
+const stemAdjectivesAndAdverbs = function( word, morphologyData ) {
+	const stemmedWord = findSuffixInClassAndStem( word, morphologyData.externalStemmer.adjectiveAndAdverbSuffixes );
 
 	if ( stemmedWord ) {
 		// Remove superlative prefix if found
-		if ( word.startsWith( "naj" ) ) {
+		if ( word.startsWith( morphologyData.externalStemmer.superlativePrefix ) ) {
```
iolandasequino, line 114: Shouldn't this function be called in the stem function as well?

packages/yoastseo/src/morphology/polish/stem.js
manuelaugustin: On a second thought, I think it makese more sense to keep it this way. Already as it is, `stemAdjectivesAndAdverbs` uses different parts of `morphologyData.externalStemmer`. `stemAdjectivesAndAdverbs` is a very broad function. In the future, it might also be expanded with e.g. exception checks. So then you'd also use data from outside `morphologyData.externalStemmer`. That's why I think limiting it to ``morphologyData.externalStemmer` doens't make too much sense.

packages/yoastseo/src/morphology/polish/stem.js
manuelaugustin: We've cleared up the misunderstanding here - the function or where it's called hasn't been changed, it's just the reference to this one string that has.

packages/yoastseo/src/morphology/polish/stem.js
manuelaugustin: Good point! I think in terms of readability (for us) it makes sense to keep in the examples in the documentation, even though we've moved the data to the morphology data file. So I'd say we can leave them in here.

iolandasequino:
I ran yarn test and yarn start and they both worked

PR #650, opened 2020-04-28
Create the DataModel component

### LINKED ISSUE #565: Create DataModel component
Parent #550

## Explanation

This issue relates to the DataModel component.

## Tasks
- [ ] Double-check with team frontend about the usecase for this component
- [ ] Create React components from the HTML
- [ ] Write tests for the components
- [x] If you come up with a better name, please change the issue and the component accordingly.

bintzandt:
## Summary
Add a DataModel component based on the provided HTML from Team Front-end ✨ 

* [@yoast/components] Adds a new DataModel component.

## Test instructions

* Check-out this branch
* Check the example components app
* Make sure that the components looks as expected 🔥 
* 🚀 

## Quality assurance

Fixes #565

### DIFF
--- a/apps/components/WordOccurrencesWrapper.js
+++ b/apps/components/WordOccurrencesWrapper.js
@@ -27,6 +27,12 @@ const initialRelevantWords = [
 	new ProminentWord( "linguïns", "linguïn", 5 ),
 ];
 
+/**
+ * A list to generate prominent words.
+ * @param {Object} props The props with relevant words
+ *
+ * @returns {React.Element} An inputlist.
+ */
 const RelevantWordInputList = ( props ) => {
 	return props.relevantWords.map(
 		( relevantWord, index ) => {
@@ -41,6 +47,12 @@ const RelevantWordInputList = ( props ) => {
 	);
 };
 
+/**
+ * A list item to generate prominent words.
+ * @param {Object} props The props with relevant words
+ *
+ * @returns {React.Element} An inputrow.
+ */
 const RelevantWordInputRow = ( props ) => {
 	function changeWord( event ) {
 		props.onChange( event.target.value, "_word", props.index );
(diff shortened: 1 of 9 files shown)


maartenleenders:
@bintzandt can you have a look at this? I've added the screenreader functions of `WordOccurrences` (which is already used in `feature/internal-linking-phase-2` on `wordpress-seo` to the `DataModel.js`, and then reused that component in `WordOccurrences.js`.
I don't think the screenreader spans will be an issue for the css/html structure.

maartenleenders:
For testing, you can check the `WordOccurrences` tab in the example app as well as the `DataModel` one.

apps/components/WordOccurrencesWrapper.js
```diff
@@ -49,7 +61,7 @@ const RelevantWordInputRow = ( props ) => {
 		props.onChange( event.target.value, "_stem", props.index );
 	}
 	function changeOccurrences( event ) {
-		props.onChange( event.target.value, "_occurrences", props.index );
+		props.onChange( parseInt( event.target.value ), "_occurrences", props.index );
```
bintzandt, line 64: Please ass the `radix` here as well. This makes sure that we will always convert the `event.target.value` to a decimal. Otherwise, it depends on the start of the string (which is something that we don't want).

```suggestion
		props.onChange( parseInt( event.target.value, 10 ), "_occurrences", props.index );
```

packages/components/src/WordOccurrences.js
```diff
@@ -48,15 +15,13 @@ const WordBarContainer = styled.li`
  */
 const WordBar = ( { word, occurrence, width } ) => {
 	return (
-		<WordBarContainer
-			barWidth={ width }
+		<li
```
bintzandt, line 18: Is the `WordBar` class still needed here?

bintzandt:
ACC: ✅ 

If the remarks from the CR are resolved this can be merged 😄  (the requested changes are not _that_ big that they require re-testing this imho)

packages/components/src/WordOccurrences.js
maartenleenders: Doh! Committed some code I was going to delete. Nice catch!

PR #666, opened 2020-05-11
Add exception check for French words ending in -s that should not be stemmed

agnieszkaszuba:
## Summary

* [yoastseo] Adds an exception check for words ending in -is/us/os where -s should not be stemmed.
* [Yoast SEO Premium] Improves keyword recognition of words ending in -is/us/os in French.

* The golden standard list was updated with the output of the new version of the stemmer (after introducing the changes), as decided in the new approach to working with golden standard lists.

## Test instructions

* Run yarn test and make sure all current specs pass and that the stemmer coverage is 100 %
* Add more examples of words that end on -is/us/os to the specs and make sure they still pass. Use examples of words that should not have the -s stemmed because they are on the exception list (sShouldNotBeStemmed list in the morphology data file), as well as words that are not on the exception list and should have the -s stemmed.
* In French Premium: try using different forms of French words that end on -us/-is/os in the keyphrase and in the title/text/meta description and see that those are recognized. Use examples both of words that are on the exception list and those that are not. For example, both _numéros_ and _numéro_  (not on the exception list) should be recognised; and both _japonais_ and _japonaise_ (on the exception list) should be recognised.

## Quality assurance

Fixes https://yoast.atlassian.net/jira/software/projects/LIN/boards/4?selectedIssue=LIN-404


manuelaugustin:
CR & acceptance 👍

PR #680, opened 2020-05-18
Add exception check for French verbs with multiple stems

agnieszkaszuba:
## Summary

* [yoastseo] Adds a check for the exception list of French verbs with multiple stems and stems them by returning the indicated canonical stem.
* [Yoast SEO Premium] Improves keyphrase recognition of certain French verbs ending on -oir or -re in the infinitive (e.g. _asseoir_, _apprendre_) as well as of some irregular verbs (e.g. _appertenir_, _vivre_)

## Test instructions

* Run yarn test and make sure all current specs pass, and that the coverage of the stemmer is 100 %
* Add more examples of forms of the verbs from the exception list (stemsThatBelongToOneWord in the morphology data file) to the specs and make sure they pass. Keep in mind that not all forms of those verbs work yet, as only the stems of the verb forms that were found while checking the golden standard list were added so far. If the particular verb form is not found on the golden standard list, it does not need to work yet.
* In French Premium: Add forms of verbs from the stemsThatBelongToOneWord list to the keyphrase and to the title/text/meta description and see that those are recognized (but see the previous point about not every form of a verb being recognised yet).

## Quality assurance

Fixes https://yoast.atlassian.net/jira/software/projects/LIN/boards/4?selectedIssue=LIN-406


manuelaugustin:
CR & acceptance 👍 

I just made some small changes: some forms ending in -s, like `appris` were on the list of `sShouldNotBeStemmed` and therefore never made it to the `canonicalizeStem` function. I've removed those from the `sShouldNotBeStemmed` function.

PR #682, opened 2020-05-20
Fix tokenization for RTL languages

manuelaugustin:
## Summary

* [yoastseo] Fixes tokenization for RTL languages. [work in progress]

* The sentence tokenizer has a check to determine whether a token has a valid sentence beginning. If that is the case, a new sentence will be started. If it's not the case, the token will be appended to the previous sentence. The general condition that determines whether a token starts with a valid sentence beginning is whether the token starts with a capital letter. Languages such as Hebrew don't use capital letters. Therefore this condition would always return `false`. This check has been modified so that a capital letter is not necessary if the letter is from a RTL script. 

* To do:
  * [ ] Apart from Hebrew, we also still need to add Arabic/Farsi/Urdu letters as a valid sentence beginning.
  * [ ] Some sanitation might be necessary, as punctuation marks aren't added in the correct position.
  * [ ] We need to test what consequences it has that we don't have the `isCapitalLetter` check in RTL languages.
  * [ ] We need to test what user-facing consequences it has that we now can tokenize sentences in RTL languages.

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #


manuelaugustin:
Closed in favor of https://github.com/Yoast/javascript/pull/735

PR #687, opened 2020-05-25
Stem French verbs with -ons suffix

agnieszkaszuba:
## Summary

* [yoastseo] Adds check for stemming -ons suffix, and for the exception list of words where only -s should be stemmed.
* [Yoast SEO Premium] Improves recognition of first person plural verb forms in French (e.g., "chantons").

* The suffix -ons is only stemmed if not preceded by -i. For most words ending in -ions, the suffix is either -ions or -s, which are covered in other steps. However, there are some words ending in -ions where -ons is the suffix. It might make sense to list those in an exception list, but I decided to not cover it in this issue because the stemming of verbs where -i precedes the suffix does not work very well in general, so it deserves a separate issue.

## Test instructions

* Make sure all current specs pass.
* Add more examples of French verb forms ending in -ons (not -ions), e.g. _achetons_, and make sure they pass. Also add more words from the exception list of words where only -s should be stemmed (nonVerbsOnOns in the morphology data file).
* In French Premium: Add French verb forms ending in -ons (not -ions), as well as other forms of the same verb (e.g., achetons - acheter), to the keyphrase and to the title/text/meta description and see if those are recognized. You can find different forms of the same verb e.g. on this site: https://conjugator.reverso.net/conjugation-french.html

## Quality assurance

Fixes https://yoast.atlassian.net/jira/software/projects/LIN/boards/4?selectedIssue=LIN-407


manuelaugustin:
CR & acceptance 👍

PR #698, opened 2020-05-28
Create reactified button components

### LINKED ISSUE #563: Create Button component
Parent #550

## Explanation

This issue relates to the Button components.

## Tasks
- [ ] Create React components from the HTML
- [ ] Write tests for the components
- [ ] There are probably a lot of buttons that are extensions from the standard button. Make sure to do an impact check of the changes that you made. If there are old buttons that we do not convert to React that you think are really useful, please contact ux or team frontend.

maartenleenders:
## Summary

* [@yoast/components] Added new standardised Buttons and Links with Yoast styling.

*

## Test instructions

* Checkout branch and load the components app.
* Check the HTML file that was provided in your browser https://github.com/Yoast/javascript/blob/b966dbd1d1ad96b8068fea9c60371ee96ee84e36/packages/components/src/button/buttons.html
* Verify that the buttons on the `Reactified components` tab look similar.
* **Note** there are still some inconsistencies, new issues will be created for the frontend devteam. 
image
* These buttons are not used in the plugin yet.
* Make sure that other props that you could potentially want to pass to a link and button (for example a `title=` to have a hovering tooltip, or an `aria-label`) are correctly passed through all the components to their eventual HTML tag. I've added the `title=` prop to the app, to demonstrate.

## Impact check

  *

## UI changes

## Quality assurance

Fixes #563


maartenleenders:
Failing snapshots are on develop. Should be checked there.

packages/components/src/button/Button.js
```diff
@@ -0,0 +1,77 @@
...
+ * @returns {ReactElemen} The Button component.
+ */
+const Button = ( props ) => {
+	// Split Button.js specific props from all other props.
+	const {
+		children,
+		className,
+		isLink,
+		...restProps
+	} = props;
```
bintzandt, line 34: You could also destructure the button props here right? Simply add `onClick` here. Would save the 5 lines below.

packages/components/src/button/Button.js
```diff
@@ -0,0 +1,77 @@
...
+
+	if ( isLink ) {
+		// Split link specific props from all additional props.
+		const {
+			href,
+			...linkProps
+		} = restProps;
```
bintzandt, line 41: Why would you destructure this if you pass all the props below anyway?

packages/components/src/button/Button.js
```diff
@@ -0,0 +1,77 @@
...
+		onClick,
+		...buttonProps
+	} = restProps;
+
+	return <button
+		onClick={ onClick }
+		className={ className }
+		{ ...buttonProps }
+	>
+		{ children }
```
bintzandt, line 63: Children can also be passed as prop.

packages/components/src/button/Button.js
```diff
@@ -0,0 +1,77 @@
...
+			href,
+			...linkProps
+		} = restProps;
+
+		return <a
+			href={ href }
+			className={ className }
+			{ ...linkProps }
+		>
+			{ children }
```
bintzandt, line 48: Children can also be passed as prop.

packages/components/src/button/Icon.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+CloseLink.propTypes = {
+	...sharedButtonPropTypes,
+	href: PropTypes.string.isRequired,
+};
+
+CloseLink.defaultProps = {
+	...sharedButtonDefaultProps,
+};
```
bintzandt, line 61: I would name the file `IconButton.js` since `Button` is also exported from `Button.js`.

One question though, why create separate `CloseLink` and `CloseButton` components if the only difference is the `isLink` prop? I think it would be nice to export one component per file, and that the user of the `CloseButton` can pass the `isLink` prop themselves. Maybe even make it required.

This is especially true, since `isLink` can still be passed as a prop to both the `CloseButton` as the `CloseLink` components. It does not do anything for `CloseLink` but `<CloseButton isLink={ true } />` renders the same as the `CloseLink`.

packages/components/src/button/Primary.js
```diff
@@ -0,0 +1,50 @@
...
+};
+
+PrimaryLink.propTypes = {
+	...sharedButtonPropTypes,
+	href: PropTypes.string.isRequired,
+};
+
+PrimaryLink.defaultProps = {
+	...sharedButtonDefaultProps,
+};
```
bintzandt, line 50: Same remarks as for the `Icon.js` apply here.

packages/components/src/button/Secondary.js
```diff
@@ -0,0 +1,50 @@
...
+};
+
+SecondaryLink.propTypes = {
+	...sharedButtonPropTypes,
+	href: PropTypes.string.isRequired,
+};
+
+SecondaryLink.defaultProps = {
+	...sharedButtonDefaultProps,
+};
```
bintzandt, line 50: Same remarks as for the `Icon.js` apply here.

packages/components/src/button/Text.js
```diff
@@ -0,0 +1,97 @@
...
+};
+
+RemoveLink.propTypes = {
+	...sharedButtonPropTypes,
+	href: PropTypes.string.isRequired,
+};
+
+RemoveLink.defaultProps = {
+	...sharedButtonDefaultProps,
+};
```
bintzandt, line 97: Same remarks as for the `Icon.js` apply here.

packages/components/src/button/Upsell.js
```diff
@@ -0,0 +1,62 @@
...
+UpsellLink.propTypes = {
+	...sharedButtonPropTypes,
+	href: PropTypes.string.isRequired,
+	caret: PropTypes.bool,
+};
+
+UpsellLink.defaultProps = {
+	...sharedButtonDefaultProps,
+	caret: false,
+};
```
bintzandt, line 62: Same remarks as for the `Icon.js` apply here.

packages/components/src/button/Icon.js
maartenleenders: Haha fair enough, but this was implemented on your recommendation 🤣 . I probably misunderstood your intentions.

I do think it's more clear for the dev to use `<CloseLink>` than <CloseButton isLink={true}>`. Especially since one is not a button, but just looks like one. But yeah, I'll remove the preset link versions.

packages/components/src/button/Button.js
maartenleenders: No because `link` don't have an `onClick` I believe.

packages/components/src/button/Button.js
maartenleenders: For legibility mostly, but I agree it doesn't add much here.

packages/components/src/button/Button.js
maartenleenders: image

eslint disagrees :(

packages/components/src/button/Button.js
maartenleenders: See above

Xyfi:
CR 👍

Xyfi:
ACC 👍

PR #713, opened 2020-06-11
Lin 434 indonesian single syllable words

FAMarfuaty:
## Summary

* [yoastseo] Adds a check for stemming Indonesian single syllable words.
* [yoastseo] N/A Adds Indonesian word form support to the content analysis app.
* [Yoast SEO Premium] Improves word form recognition of Indonesian single syllable words.

* This PR only concerns Indonesian single syllable words with this form [di-] + single syllable word + [-kan/-i] + [-ku/-mu/-nya] + [-kah/-lah/-pun], *[ ] being optional. For the single syllable words that get prefix penge-/menge-, the check will be added in different issue.
* However, as the result of implementing this check, two syllable words whose stem starts with di- or ends in ending that look like a valid suffix like -i are overstemmed. One solution for this to have a check before stemming single syllable words whether the input word is indeed a word that should not be stemmed.

## Test instructions

Check the unit tests:

* Run yarn test and make sure all tests still pass.
* Add more single syllable word forms in the specs and make sure that they have the same stem. E.g. diklik, diklikkan, klikkan, kliklah -> klik. Other example of single syllable words can be found here https://github.com/Yoast/YoastSEO.js-premium-configuration/pull/186

Check in the content-analysis app:

* Open the content analysis app by running yarn start;
* Insert a text of al least 150 words;
* Set the locale to Italian (i.e., id_ID);
* Make sure you have different forms of single syllable words you want to test in the text and set the word
you want to test as a keyword (e.g., vasmu, ditik, dicekkan, balnya );
* Check on the eye icon that all the word forms become highlighted 

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LIN-434


packages/yoastseo/src/morphology/indonesian/stem.js
```diff
@@ -165,6 +230,12 @@ const stemDerivational = function( word, morphologyData ) {
  * @returns {string} The stem of Indonesian word.
  */
 export default function stem( word, morphologyData ) {
+	const singleSyllableWords = stemSingleSyllableWords( word, morphologyData );
+	// Stem the single syllable words
+	if ( singleSyllableWords ) {
```
nataliashitova, line 235: The `stemSingleSyllableWords` function is created in such a safe way that it always returns a valid word - either a single-syllable word from the list or the original word. The `removeEnding` function called within it also returns safe. I think you can simplify the code above by running `stemSingleSyllablWords` always and reassigning word to its result.

FAMarfuaty: If we use the result as it is without discarding the result when it has more than 1 syllable, there are many words that are going to be incorrectly stemmed, especially words that after stemming prefix di- have a beginning that looks like a valid prefix ( e.g. ter- in (di-) terjemah). And I think it's safer to ignore the result when it has more than 1 syllable unless we have an exception list of words that have a beginning that looks like a valid prefix.

PR #718, opened 2020-06-12
Have less buttons and manage more via props

maartenleenders:
## Summary

* Yoast styled buttons are now available in preset variants, such as `primary`, `secondary`, `buy`, etc.
* Yoast styled links (identically styled to the buttons) are now available in preset variants, such as `primary`, `secondary`, `buy`, etc.

* RTL styling is off, not fixed in this PR.
* `SvgIcon` was not suited to work for us here, especially given the workflow `team-frontend` desires.
* Adding icons to buttons works by setting them in CSS as an image mask on distinct class. Then, you add a variant name to the `variantToIcon` object near the top of `Button.js` like so: 

```js
variantName: { iconAfter: "yoast-button--classNameOfSpanWithMask" }
```
(use `iconBefore` if you need the span with the icon to appear in front of the text ).

* Add variants is as simple as adding the variant name as a key to `iconToClassName` and setting the desired class as a value. The often occurring `yoast-button yoast-button--` is available as a constant.

## Test instructions

* Checkout the components app, and verify that all button and links variants look and work as they should.
* Perhaps try to add some variants yourself, to see if it "feels" nice to work with them.
* Note that RTL is (I think) erroneously implemented, so that needs some work once we get to these buttons in the future.

## Quality assurance

Fixes #


maartenleenders:
I am running into limitations wrt our "new" approach for icons (by setting a className on a span):

```
.yoast .yoast-button--buy__caret {
	flex-shrink: 0;
	display: inline-block;
	width: 6px;
	height: 16px;
	/*Negative margin to compensate the icon white space within its viewbox.*/
	margin: 0 -2px 0 6px;
	background: var(--yoast-svg-icon-caret-right) center center no-repeat;
}
```

Ideally, I would like to be able to tell the `Button` component which icon I want (i.e. have control over the `var( --yoast-svg-icon-caret-right )` part), without having to create write a new CSS class for a span.

Wouldn't it be better to have something like a `yoast-button--icon__span` that can hold any svg and styles it appropriately?

That way I can just render 
```js
<span className=".yoast-button--icon__span">
	<SvgIcon icon={ props.theDesiredIcon } /> // This I can swap out based on props
</span>
```

maartenleenders:
Decision: We're going with the `<span>` option anyways, swapping out the span's className based on the props.

maartenleenders:
> I was wondering: what is the reason that the `CloseButton` does have its own file?
> 
> Isn't it better to move that into the `Button.js` file as well?

I have kept it separate because it doesn't follow the `potential icon before`-`text`-`potential icon after` logic.

It's just an icon that you can click. Probably best to call it a `ClickableIcon` rather than `CloseButton` but 🤷

bintzandt:
CR: One question remains to be answered.
ACC: :heavy_check_mark:

PR #721, opened 2020-06-16
Add a private key and certificate so that openssl does not complain

bintzandt:
On my Linux machine, OpenSSL gave the following error: Error: error:140AB18F:SSL routines:SSL_CTX_use_certificate:ee key too small.

The problem is that the default stubby certificates use a RSA key with size 1024 which OpenSSL considers to be too small (although for development there is no real risk).

To circumvent this problem, this PR adds a selfsigned key and certificate with RSA 2048 keys. This is enough to make OpenSSL happy again.

## Summary

* [example-components-app] Adds a private key and a certificate to the example components app to make it work on Linux.

## Test instructions

* Test that the example components app still works on Mac.

## Quality assurance

Fixes N/A


bintzandt:
Build is failing because the base branch is out-of-date (regarding `getMorphologyData`). I do not think that it is necessary to fix that per se.

Will be fixed once develop is merged in the feature branch.

PR #724, opened 2020-06-17
Improve stemming single syllable words

FAMarfuaty:
## Summary

* [yoastseo] Adds a check for stemming Indonesian single syllable words.
* [Yoast SEO Premium] Improves word form recognition of Indonesian single syllable words.

* This PR only concerns Indonesian single syllable words with this form **[menge-] + single syllable word + [-kan/-i]** and **[penge-] +single syllable word + [-an]**, with [] being optional. 

* However, as the result of implementing this check, two syllable words whose stem starts with one of the words in singleSyllableWords list and ends with -kan/-an/ are going to be overstemmed. E.g. lokan -> lo ("lo" is in the list)

* Two syllable words whose stem starts with ke- and ends in one of the single syllable words which gets prefix meng- ( **meng-/peng- + ke + single syllable word + [-kan/-an/-i]**  ). Example, **mengebiri**  is stemmed into **bir** (single syllable word) while it actually should be stemmed into **kebiri** (“ebiri” is listed in kBegining list)
Other example  **mengecap** / **pengecap** is currently stemmed into **cap** which is a valid word, but it can also be stemmed into **kecap** ("ecap" is in the kBeginning list).

## Test instructions

Check the unit tests:

* Run yarn test and make sure all tests still pass.
* Add more single syllable word forms in the specs and make sure that they have the same stem. E.g. pengeklik, mengeklik, pengeklikan, mengeklikkan, mengekliki, klikan, klikkan -> klik
pengecekan, pengecek, mengecek, mengecekkan, mengeceki, cekkan -> cek. Other example of single syllable words can be found here Yoast/YoastSEO.js-premium-configuration#186 and add the possible affixes, see pattern above.
* Also see the relevant technical choice for an overview of words that are not going to stemmed correctly.

Check in the content-analysis app:

* Open the content analysis app by running yarn start;
* Insert a text of al least 150 words;
* Set the locale to Indonesian (i.e., id_ID);
* Make sure you have different forms of single syllable words you want to test in the text and set the word you want to test as a keyword (e.g., pengetesan, pengetes, mengetes, mengeteskan, mengetesi, teskan, tes, dites, tesnya, teslah, tesku, teskanlah should be all recognised );
* Check on the eye icon that all the word forms become highlighted

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LIN-440


packages/yoastseo/src/morphology/indonesian/stem.js
```diff
@@ -34,13 +34,33 @@ const checkSingleSyllableWordSuffix = function( word, suffixesArray ) {
...
+ *
+ * @param {string}	word			The word to check.
+ * @param {Object}	morphologyData	The Indonesian morphology data file.
+ *
+ * @returns {string} The stemmed word.
+ */
+const stemSingleSyllableWordsPrefixes = function( word, morphologyData ) {
+	// If the word gets prefix di-, stem the prefix here. E.g. dicekkanlah -> cekkanlah, dibomi -> bomi
+	if ( word.startsWith( "di" ) && checkBeginningsList( word, 2, morphologyData.stemming.singleSyllableWords ) ) {
+		word = word.substring( 2, word.length );
```
manuelaugustin, line 48: Is it possible that a word has both `di` and `menge/penge`? For example `dimengesomthing`? Judging from the issue description, that doesn't seem to be the case. However you still check `menge/penge` even if the `di` check results in stemming. You can improve that by having a `return` within the `if` statements for the two checks, because then you'll return the word right away if the `if` statement holds true.

packages/yoastseo/src/morphology/indonesian/stem.js
```diff
@@ -70,7 +87,7 @@ const stemSingleSyllableWords = function( word, morphologyData ) {
 			morphologyData.stemming.doNotStemWords.doNotStemPronounSuffix, morphologyData );
 
 		// If the word ends in -kan/-i suffix and has exactly 2 syllables, stem the suffix. E.g. cekkan -> cek, bomi -> bom
-		if ( /(kan|i)$/i.test( word ) && calculateTotalNumberOfSyllables( word ) === 2 ) {
+		if ( /(kan|an|i)$/i.test( word ) && calculateTotalNumberOfSyllables( word ) === 2 ) {
```
manuelaugustin, line 90: I'm wondering if there isn't a way to avoid some of the overstemming issues you're mentioning, specifically cases like `turban`, where `tur` is in the list.

For the `kan/an/i` check, we check whether the word being checked ends with one of those suffixes. But can't we check at this point whether the word is one of the words from the single syllable words list + one of those suffixes? Since we only have `tur` in the list, and not `turb`, for cases like `turban` it should avoid overstemming. (It doesn't work for cases like `lokan`, because there `lo` + `kan` would be matched.)

Are there reasons not to do that? Would we understem some words in that case?

packages/yoastseo/src/morphology/indonesian/stem.js
FAMarfuaty: Yes you're right. I don't think it's possible to have `dimengesomthing` pattern. And adding the return statement within `if` block is a good idea!

packages/yoastseo/src/morphology/indonesian/stem.js
FAMarfuaty: I guess I missed this. I tried to check at that point whether the word is one of the words from the single syllable words list + one of those suffixes. And the results are so much better!

manuelaugustin:
CR & acceptance 👍

PR #754, opened 2020-07-02
FRO-209 table component

rolf-yoast:
## Summary

* Introduce the new table component

* N/A

## Test instructions

* Check out this PR
* Check if the standard table components are okay and inline with this design https://www.sketch.com/s/1ede3372-d1cc-4803-a4af-a06ca86aa0e3/a/ZAl4Yv#Inspector

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/FRO-209


packages/components/src/table/table.css
```diff
@@ -0,0 +1,48 @@
...
+.yoast-table th {
+	color: var(--yoast-color-dark);
+	text-align: left;
+	padding: 18px 12px;
+}
+
+.yoast-table td {
+	padding: 18px 12px;
+	border-bottom: var(--yoast-border-default);
+	font-size: var(--yoast-font-size-default);
```
JoseeWouters, line 33: I think the font-size is inherited from the parent, no need to set it here?

packages/components/src/table/table.css
```diff
@@ -0,0 +1,48 @@
...
+	width: 100%;
+}
+
+.yoast-table tfoot th {
+	border-bottom: none;
+}
+
+.yoast-table th,
+.yoast-table tr {
+	border-bottom: var(--yoast-border-default);
```
JoseeWouters, line 17: The `border-bottom` is set on the `td` too, so I'd remove this ruleset and just add the border for the `th` on line 24

packages/components/src/table/table.css
```diff
@@ -0,0 +1,48 @@
...
+	padding: 18px 12px;
+}
+
+.yoast-table td {
+	padding: 18px 12px;
+	border-bottom: var(--yoast-border-default);
+	font-size: var(--yoast-font-size-default);
+}
+
+.yoast-table .yoast-td--right {
```
JoseeWouters, line 36: This classname is not really clear, maybe we could use a more generic class for text-alignment or maybe target only the second td.

packages/components/src/table/table.html
```diff
@@ -0,0 +1,89 @@
...
+					</a>
+				</th>
+				<th>Trend</th>
+				<th>Action</th>
+			</tr>
+		</thead>
+		<tbody>
+			<tr>
+				<td>yoast seo plugin</td>
+				<td class="yoast-td--right">800</td>
```
JoseeWouters, line 63: These classes can be removed too.

PR #710, opened 2020-06-09
P3-32 Adjust styling for the related keyphrases modal

afercia:
## Summary

This PR can be summarized in the following changelog entry:

* Adjusts the styling provided for the modal component to use it for the Gutenberg Modal component.

Note: this PR is needed for https://github.com/Yoast/wordpress-seo/pull/15383
Note 2: the name of this PR branch starts with P3-5 instead of P3-32, please ignore it

## Relevant technical choices:

- adds a few class names to target the Gutenberg Modal elements for general styling
- adjusts the modal header and footer CSS to match the design: accordingly to the design team feedback:
  - on desktop:
    - header height: 72px
    - footer height: 88px
  - on mobile:
    - header height: 48px 
    - footer height: 72px
- changed indentation to use tab characters
- added the missing `--yoast-` prefix to the CSS variables in the `modal.css` file

This PR does _not_ fix other issues in the modal CSS. For example, the `modal-react-tabs.html` modal content has some layout issues on a viewport width less than 782px. This is a pre-existing problem, fixing it is out of the scope of this PR. Screenshot of the problem:

modal-react-tabs on develop

## Test instructions

This PR can be tested by following these steps:

Check the 3 modal demos by opening the following files in your browser:
1. packages/components/src/modal/modal-without-nav.html
2. packages/components/src/modal/modal-react-tabs.html
3. packages/components/src/modal/modal-mobile-collapsible.html

- check that on a viewport width less than 782 pixels the header and footer heights are smaller.
- check that the modals content is scrollable
- check there are no unwanted changes 
- note that the modal 3 works only on a small viewport width (I guess this is the expected behavior for this demo)
- as mentioned above, there are pre-existing issues that are out of the scope of this issue

## Impact check

* This PR affects the following parts of the plugin, which may require extra testing:
  *

## UI changes
* [ ] This PR changes the UI in the plugin. I have added the 'UI change' label to this PR.

## Quality assurance

* [x] I have tested this code to the best of my abilities
* [ ] I have added unittests to verify the code works as intended

Fixes [P3-32]

PR #771, opened 2020-07-07
Revert p3 5 create related keyphrases component

igorschoester:
## Summary

* Reverts https://github.com/Yoast/javascript/pull/710

*

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #


igorschoester:
The `modal.css` file is the only thing that is actually loaded.
The `yoast-tabs` class is the only one that is not prefixed with a modal part, and that is not actually used anywhere.
So I think this revert is not actually needed. Would save some work if we leave it.

PR #836, opened 2020-08-20
Socials: change width to max-width

igorschoester:
This makes it so the previews fit on smaller screens.

## Summary

* Fixes a bug where the social previews would be able to break out of their container when viewed on smaller screens.

* Removed some extra constrains on the height, as it was not listening to the hidden description. The height only makes sense on the image itself.
* Made the description hide when the screen width is smaller than the image width. Not perfect, but seems to work fine.

## Test instructions

* Combine with premium.
* Go to the social previews.
* Switch to a smaller screen (iPhone 5 or something?).
* Verify that the previews no longer break out.
* Verify that the descriptions for both Facebook and Twitter are hidden when the screen is smaller than the wanted image width.
* Change the `SEO -> Social -> Twitter -> The default card type to use` to `Summary` instead of `Summary with large image`.
* Verify that the image is unchanged.

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/P1-141

### DIFF
--- a/packages/social-metadata-previews/src/facebook/FacebookPreview.js
+++ b/packages/social-metadata-previews/src/facebook/FacebookPreview.js
@@ -61,7 +61,6 @@ const FacebookPreviewWrapper = styled.div`
 	flex-direction: ${ props => props.mode === "landscape" ? "column" : "row" };
 	background-color: #f2f3f5;
 	max-width: 527px;
-	height: ${ props => determineWrapperHeight( props.mode ) };
 `;
 
 const FacebookTextWrapper = styled.div`
@@ -78,7 +77,6 @@ const FacebookTextWrapper = styled.div`
 	flex-direction: column;
 	justify-content: ${ props => props.mode === "landscape" ? "flex-start" : "center" };
 	max-width: ${ props => determineTextContainerWidth( props.mode ) };
-	${ props => props.mode === "landscape"  ? "height: 78px;" : "" }
 	font-size: 12px;
 	overflow: hidden;
 `;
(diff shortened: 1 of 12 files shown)


packages/social-metadata-previews/src/facebook/FacebookPreview.js
```diff
@@ -42,7 +42,7 @@ const determineWrapperHeight = ( mode ) => {
 const determineTextContainerWidth = ( mode ) => {
 	switch ( mode ) {
 		case "landscape":
-			return "527x";
+			return "527px";
```
bintzandt, line 45: Lol :sweat_smile:

bintzandt (approved):
CR :+1:

packages/social-metadata-previews/src/facebook/FacebookPreview.js
```diff
@@ -60,8 +60,7 @@ const FacebookPreviewWrapper = styled.div`
 	display: flex;
 	flex-direction: ${ props => props.mode === "landscape" ? "column" : "row" };
 	background-color: #f2f3f5;
-	width: 527px;
-	height: ${ props => determineWrapperHeight( props.mode ) };
```
bintzandt, line 64: Why did you remove this?

bintzandt:
Your PR looks fine. The Social Previews behave much better.

However, the metabox still does not look nice. But that is outside the scope of this issue.

Screenshot is of an "iPhone 6/7/8" running iOS 11.

PR #839, opened 2020-08-25
Replace Select2 with react-select for our multiselect

bintzandt:
## Summary
This PR changes the `MultiSelect` implementation of the components package to use `react-select` instead of `Select2`

* [@yoast/components] Changes the `MultiSelect` implementation to use `react-select` instead of `Select2`.

* Styling has been done using CSS to keep everything aligned with how the rest of the new components are styled.
* `yoast-select-container` has been prefixed to all CSS selectors to increase specificity.
* Removed the snapshot test for the `MultiSelect` since the `useCallback` hook was causing problems with` react-test-renderer`. I spend some time investigating but I could not reach a solution. Since we are removing snapshot tests anyway in the plugin, I thought it would be okay to remove it here as well.

	This was the error for those interested.
	```
	Invalid hook call. Hooks can only be called inside of the body of a function component. This could 	happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

      80 |      const selectedOptions = reactSelectOptions.filter( option => selections.includes( option.value ) );
      81 | 
    > 82 |      const onChangeHandler = useCallback( selection => {
      83 |              // Only call the onChange handler on the selected values.
      84 |               onChange( selection.map( option => option.value ) );
      85 |      } );
	```
      

## Test instructions

### Example components app
1. Start the example components app
2. Go to the reactified components tab
3. Checkout the MultiSelect
4. Verify that the component looks as expected.
5. Verify that changing the options in the JS changes the way the component is rendered.
**NOTE: Changing the value of the select using the Dropdown DOES NOT WORK. The component needs to be managed via an external store to set the props correctly. This is not implemented in the example components app.**
### WordPress SEO
1. Make sure that you are on the `P1-147-replace-select2-with-react-select` in WordPress SEO
2. Link the monorepo and build everything
3. Check out the Advanced Settings component in the metabox
4. Verify that the styling aligns with that of the example components app.
5. Change the value of the hidden input (`#yoast_wpseo_meta-robots-adv`) and save the page.
	* Possible values are `nosnippet`, `noarchive` `noimageindex` or a combination of those separated with a `,`.
6. Reload and verify that the MultiSelect renders with the correct values.
**NOTE: Changing the value of the select using the Dropdown DOES NOT WORK. The component needs to be managed via an external store to set the props correctly. This is implemented in https://github.com/Yoast/wordpress-seo/pull/15917.**

## Impact check

  *

## UI changes

## Quality assurance

Fixes P1-147


igorschoester (approved):
CR 👍 

I did a small `useState` experiment, but the example app did not like that.

bintzandt:
> CR +1
> 
> I did a small `useState` experiment, but the example app did not like that.

Yeah, I also tried and failed...

igorschoester:
Acceptance ✅

PR #840, opened 2020-08-25
Let prominent words threshold be dependent on the availability of morphology data

johannadevos:
## Summary

* Sets the threshold for words to be recognized as prominent words from 4 to 2, in case a language has no morphology support. This increases the chance that different word forms are saved, leading to better internal linking suggestions for these languages.

*

## Test instructions

**Test in a language with morphology support**
* In `Settings`, set your site language to one that has morphology support, for example English.
* Open a post (either new or existing).
* Open the `Network` tab in the inspector.
* Add (English) content to the post. 
* In the `Network` tab, see a request starting with `link_suggestions?prominent_words`.
* Click on the request, then `Headers`. Scroll down and see the prominent words. 
* See that the lowest value a prominent word can have, is 4. 
* This may require some fiddling with your content. 
    * Make sure that you have some words/word forms that occur exactly 4 times, and some words/word forms that occur 3 times (the latter ones, you shouldn't see in the request).
    * Make sure that your text isn't too long. Only the 20 prominent words are shown that occur most frequently, so you don't want to have 20 or more words/word forms with 5 or more occurrences.
    * Make sure that your text is at least 100 words, otherwise no prominent words will be generated.

**Test in a language without morphology support**
* In `Settings`, set your site language to one that has morphology support, for example Danish.
* Open a post (either new or existing).
* Open the `Network` tab in the inspector.
* Add (Danish) content to the post. 
* In the `Network` tab, see a request starting with `link_suggestions?prominent_words`.
* Click on the request, then `Headers`. Scroll down and see the prominent words. 
* See that the lowest value a prominent word can have, is 2. 
* This may require some fiddling with your content. 
    * Make sure that you have some words/word forms that occur exactly 2 times, and some words/word forms that occur 1 time (the latter ones, you shouldn't see in the request).
    * Make sure that your text isn't too long. Only the 20 prominent words are shown that occur most frequently, so you don't want to have 20 or more words/word forms with 5 or more occurrences.
    * Make sure that your text is at least 100 words, otherwise no prominent words will be generated.

## Impact check

  * The calculation of prominent words.

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/P2-138

### DIFF
--- a/packages/yoastseo/spec/researches/getProminentWordsForInternalLinkingSpec.js
+++ b/packages/yoastseo/spec/researches/getProminentWordsForInternalLinkingSpec.js
@@ -25,7 +25,7 @@ describe( "relevantWords research", function() {
 		expect( words ).toEqual( expected );
 	} );
 
-	it( "does return prominent words for texts with more than 300 words", function() {
+	it( "returns prominent words for texts with more than 300 words", function() {
 		const paper = new Paper( "texte" + " et texte".repeat( 180 ), { title: "Title" } );
 
 		const researcher = new Researcher( paper );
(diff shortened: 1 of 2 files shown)


johannadevos:
Please also give feedback on the changelog entry, I'm not sure whether it should contain the phrase "prominent words", but I also don't know a good alternative at this point.

hansjovis (approved):
CR: 👍

hansjovis:
> Please also give feedback on the changelog entry, I'm not sure whether it should contain the phrase "prominent words", but I also don't know a good alternative at this point.

Perhaps we can add a line telling the users why we made this change:
> Sets the threshold for words to be recognized as prominent words from 4 to 2, in case a language has no morphology support. This increases the chance that different word forms are saved, leading to better internal linking suggestions for these languages.

hansjovis:
Acceptance test: ✅

PR #848, opened 2020-08-28
Use value rather than default value so the component can be controlled

maartenleenders:
## Summary

Because the AdvancedSettings Tab is now in two places, their values can go out of sync. We need a way to update the TextInput and TextArea components with external props from e.g. a redux store.

Therefore they need to be switched to be controlled components.

* TextInput and TextArea have been changed to controlled components, their value should be passed from a higher order component.

* 

## Test instructions

* Checkout `feature/metabox-modal` on free or premium, and `yarn link-monorepo` to this branch.
* Scroll down to the metabox, and open the advanced settings tab. Make sure you see the CanonicalURL input.
* Now open the modal, open its AdvancedSettings, and change the CanonicalURL input. Note that in the background, the advancedsettings in the Metabox are updated accordingly.
* Also test the components on their own, in the monorepo. Check the components app.

## Impact check

  *

## UI changes

## Quality assurance

Fixes P1-66

### DIFF
--- a/packages/components/src/inputs/TextArea.js
+++ b/packages/components/src/inputs/TextArea.js
@@ -24,7 +24,7 @@ const TextArea = ( props ) => {
 			<textarea
 				id={ id }
 				name={ props.name }
-				defaultValue={ props.value }
+				value={ props.value }
 				className="yoast-field-group__textarea"
 				aria-describedby={ props.ariaDescribedBy }
 				onChange={ props.onChange }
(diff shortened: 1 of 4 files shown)


bintzandt:
It seems that these components are only used inside the AdvancedSettings tab. That means that this change will not break behavior anywhere else.

bintzandt (approved):

bintzandt:
ACC: :+1:

PR #864, opened 2020-09-18
P2 214 add valid function

increddibelly:
## Summary

* Added "Valid" method to abstract base classes.

* Return true in the base classes; subclasses may add own validation (see e.g. P2-212, P2-213)

## Test instructions

* The method is not used yet, the new code is harmless. Cannot be tested until subclasses implement their own validation or innerblock validation.

## Impact check

  * none.

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ This monorepo includes apps for testing purposes. These apps are not published. 
 The following commands can be executed from the javascript project root:
 
 * `yarn install`, will install all dependencies for all packages.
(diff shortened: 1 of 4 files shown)


hansjovis (approved):
CR: 👍 

Code looks good to me.
Travis build fails on a known issue in the feature branch, so does not block merging.

I do find it odd that code changes that are already in the feature branch are added in this commit as well 🤔. However, since GitHub does not indicate any problems I am going to merg this.

PR #865, opened 2020-09-21
Add prototype assessment for ranking intentions

iolandasequino:
## Summary

*  [yoastseo] Adds a prototype assessment to recognise ranking intentions in English. 
*  [Yoast SEO Premium] Adds a prototype assessment to recognise ranking intentions in English based on the ratio between singular and plural keyphrase occurrences in the text.

* The assessment does not currently work on cornerstone articles, taxonomy articles and for related keyphrases.

## Test instructions

* Set your site to English;
* Test in Premium;
* Paste a text of at least 150 words;
* Set a singular keyphrase, e.g., `fluffy cat`;
* Add a number of occurrences of your singular keyphrase to the text;
* Add the same number of occurrences of the plural form of your keyphrase, i.e., `fluffy cats`;
* The ranking intention assessment should turn orange and read: `Your text does not reflect any particular ranking intention.  If your keywords is singular, use more singular occurrences; if your keyphrase is plural, use more plural occurrences!`;
* Change some of the singular occurrences to plural until the ranking intention bullet turns red and reads: `Your text does not reflect your ranking intention. Change your keyphrase occurrences!`;
* Change some of the plural occurrences to singular until the ranking intention bullet turns green and reads: `Your text reflects your ranking intention. Good job!`;
* Check that the same holds for a plural keyphrase;

* Check that the site does not crash and that no bullet is there for the assessment when you are in Free.
* Check that the assessment isn't shown when the site is set to another language than English.

Extra internal test instructions:
* Run yarn test and make sure all tests pass

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LIN-544


packages/yoastseo/spec/assessments/SingularPluralAssessmentPrototypeSpec.js
```diff
@@ -0,0 +1,90 @@
...
+		expect( result.getScore() ).toBe( 9 );
+		expect( result.getText() ).toBe( "<a href='https://yoa.st/33v' target='_blank'>Ranking intention</a>:" +
+			" Your text reflects your ranking intention. Good job!" );
+	} );
+	it( "runs the ranking intention on the paper with keyword", function() {
+		const paper = new Paper( "There is more than a pot you can choose for your plant. The pot with tribal pattern is our bestseller.",
+			{ keyword: "plant pots" } );
+		const researcher = new Researcher( paper );
+		const result = new SingularPluralAssessmentPrototype().getResult( paper, researcher, i18n );
+		expect( result.getScore() ).toBe( 6 );
```
manuelaugustin, line 34: The scores changed a bit because we now use the average of all keyphrase form percentages. Could you add a spec for a score of 3?

manuelaugustin (changes requested):
I've left some comments and also changed some things in the code.

What I've changed:
- Use the average of all percentages of original/modified forms to calculate a result. Originally, only the percentage of the last words was used. So for `pattern plant pots` only the percentage for `pots` vs. `pot` would be used, while the previous results were incorrectly overwritten.
- I changed the markings so that all modified occurrences would be marked, as it's those that need to be changed.
- I introduced a gray bullet for when there are no keyphrase occurrences in the text.
- I changed the internal structure of the research to make the code a bit more readable (see comment in commit).
- I added some specs for the `isApplicable` conditions.

packages/yoastseo/src/assessments/seo/SingularPluralAssessmentPrototype.js
```diff
@@ -0,0 +1,198 @@
...
+	 * Calculates the percentage of the occurrences of the singular and plural forms in the text.
+	 *
+	 * @returns {number}    The percentage of the occurrences of the singular and plural forms in the text.
+	 */
+	determinePercentage() {
+		const originalModifiedPairs = this.originalModifiedPairs;
+		const percentages = [];
+
+		// Prevent division by zero errors.
+		if ( originalModifiedPairs.length !== 0 ) {
```
manuelaugustin, line 79: After the changes I made in the code this isn't enough to provide division by zero changes. 

You could use the same function that's used to calculate `noKeyphraseOccurrences` to determine that `originalCount` and `modifiedCount` aren't both 0.

manuelaugustin:
On thing I forgot to mention:

In principle, the assessment should also be added to the following assessors:
cornerStoneAssessor, taxonomyAssessor,  relatedKeywordAssessor, relatedKeywordTaxonomyAssessor, relatedKeywordTaxonomyAssessor

I don't think it's necessary for the prototype, but it's good to mention in the relevant technical choices that the assessment does not currently work on cornerstone articles, taxonomy articles and for related keyphrases.

manuelaugustin:
CR 👍

I just made some small changes, see my latest commit.

Acceptance 👎

To problems:

1)
You already mentioned that with the mean, the result isn't as good. However, I noticed that now it's actually impossible to get a bad result with the testing instructions provided. I think you're right that we need to change the calculation of the result.

One idea I had is this: instead of returning the mean of all percentages, we could return the lowest percentage (i.e. `Math.min( ...percentages )` instead of `mean( percentages )`).

I tried that out and it made the results a bit better I think. But maybe in that case the boundaries are a bit too strict. Maybe you can have a look into that?

2) Originally, the assessment would return an error when you run it in Free. That's because no morphology data is available in that condition. The functionality used in the assessment requires morphology data, however. I've now added a check for whether morphology data is available. The assessment won't be run now if no morphology data is available. Could you also add testing instructions for that?

manuelaugustin:
Final CR & acceptance 👍

PR #878, opened 2020-10-06
P3-184 changes for zapier ui

afercia:
## Summary

[components] Add styles for the input field and button for the Zapier integration.

To match the design for the Zapier integration the following adjustments and new styles are needed:
- The box-shadow of the button should be `box-shadow: inset 0 -2px 0 rgba(0, 0, 0, 0.1);`
- A grey background for the input field when it's readonly or disabled.
- A style variant to make the input label element not bold.
- Horizontal alignment of input field + button: the current CSS makes the input field full width so the button goes in a new line; we need styling to make the input + button horizontally aligned

Design to match:

Screenshot 2020-10-06 at 11 41 30

See https://yoast.atlassian.net/browse/P3-184

## Test instructions

- open the `packages/components/src/inputs/input.html` file in your browser and see the new examples

## Impact check

  * Zapier integration tab

## UI changes

## Quality assurance

See https://yoast.atlassian.net/browse/P3-184

### DIFF
--- a/packages/components/src/button/buttons.css
+++ b/packages/components/src/button/buttons.css
@@ -7,7 +7,7 @@
 	text-decoration: none;
 	border-radius: 4px;
 	border: 1px solid rgba(0, 0, 0, 0.2);
-	box-shadow: inset 0 -2px 0 rgba(0, 0, 0, 0.3);
+	box-shadow: inset 0 -2px 0 rgba(0, 0, 0, 0.1);
 	transition: background-color 150ms ease-out 0s;
 	line-height: 1.2;
 	font-size: 14px;
--- a/packages/components/src/button/buttons.css
+++ b/packages/components/src/button/buttons.css
@@ -7,7 +7,7 @@
 	text-decoration: none;
 	border-radius: 4px;
 	border: 1px solid rgba(0, 0, 0, 0.2);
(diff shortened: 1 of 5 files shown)


afercia:
Note: for now, I kept the new styles strictly coupled to the specific usage for the Zapier design. For example the horizontal alignment takes in consideration only the specific case of an input field followed by a button. In the future, this could be improved to cover also other cases but for now we don't need anything else for Zapier.

packages/components/src/button/buttons.css
```diff
@@ -7,7 +7,7 @@
 	text-decoration: none;
 	border-radius: 4px;
 	border: 1px solid rgba(0, 0, 0, 0.2);
-	box-shadow: inset 0 -2px 0 rgba(0, 0, 0, 0.3);
+	box-shadow: inset 0 -2px 0 rgba(0, 0, 0, 0.1);
```
rolf-yoast, line 10: This is incorrect. This should stay 0.3. Only the secondary button should have 0.1

rolf-yoast (changes requested):

rolf-yoast:
CR 👍

rolf-yoast (approved):

PR #888, opened 2020-10-20
Refactor the keyphraseDistribution research

agnieszkaszuba:
## Summary

* [yoastseo] Creates language-specific keyphraseDistribution researches

* The assessment applies different scoring criteria depending on whether the language has function words support or not. That's why the function in the basic research takes a boolean value that tells it whether the language has function words. The value is false in the default research and true in the language-specific researches for languages with function words. However, this means that every language-specific research looks identical (as opposed to other language-specific researches where usually every language is different). Maybe there is a more efficient way to handle this in the future so that there is no need for so many identical files?

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LINGO-477


packages/yoastseo/src/languageProcessing/languages/de/researches/keyphraseDistribution.js
```diff
@@ -0,0 +1,12 @@
...
+import getFunctionWords from "../config/functionWords";
+const functionWords = getFunctionWords().all;
+
+/**
+ * Determines which portions of the text did not receive a lot of content words from keyphrase and synonyms.
+ *
+ * @inheritDoc keyphraseDistributionResearcher
+ */
+export default function( paper, researcher ) {
+	return keyphraseDistributionResearcher( paper, researcher, functionWords );
```
omarreiss, line 11: You pass true or false everywhere else. Here you pass a list. Is this intentional?

packages/yoastseo/src/languageProcessing/researches/base/keyphraseDistribution.js
```diff
@@ -182,10 +178,11 @@ const getSentenceScores = function( sentences, topicFormsInOneArray, locale ) {
  *
  * @param {Paper}       paper       The paper to check the keyphrase distribution for.
  * @param {Researcher}  researcher  The researcher to use for analysis.
+ * @param {boolean}     funcWords   The function words list.
  *
  * @returns {Object} The scores of topic relevance per portion of text and an array of all word forms to highlight.
  */
-const keyphraseDistributionResearcher = function( paper, researcher ) {
+const keyphraseDistributionResearcher = function( paper, researcher, funcWords ) {
```
omarreiss, line 185: `funcWords` implies an array. What would be a parameter name that describes the boolean well?

omarreiss (changes requested):
Looking alright. Two things that need fixing.

packages/yoastseo/src/languageProcessing/languages/de/researches/keyphraseDistribution.js
FAMarfuaty: That wasn't intentional. We just fixed it.

packages/yoastseo/src/languageProcessing/researches/base/keyphraseDistribution.js
FAMarfuaty: We changed it to `hasFunctionWords` in the latest commit

PR #916, opened 2020-11-04
P2-354 Set up typescript tests for the `schema-blocks` package.

hansjovis:
## Summary

* Sets up typescript tests for the `@yoast/schema-blocks` package. 

* Decided to setup the Jest configuration in a separate file instead of in `package.json` like the other packages.
  * This makes the Jest configuration more easily discoverable (file in the root folder, next to the other configuration files) instead of hidden in the `package.json`.
  * Bonus points: comments.
* Had some problems with making the latest `jest` version work with the latest `ts-jest` version.
  * Ultimately opted for the combination of versions [used in the MyYoast server](https://github.com/Yoast/my-yoast/blob/master/nestjs-server/package.json), which are also TypeScript based.
* Decided to add the documentation to a `README.md` in the package.
  * This way the documentation lives as close to the code as possible, which should make it easily discoverable.
  * GitHub will show it when looking at the package / folder on GitHub.

## Test instructions

* Run `yarn test` in the root of the `@yoast/schema-blocks` package (e.g. `/packages/schema-blocks`).
  * This will run the single test that has been added in this PR.
* The tests should pass.
* Run `yarn test --coverage`.
  * This will run the same tests, but now with coverage enabled.
* The tests should pass.
* The coverage of the test should be 100%.

## Impact check

  * Considering this is a monorepo and JavaScript dependencies are hoisted up to the root, possible Jest version conflicts may arise. 
    * Run `yarn test` in the root of the monorepo to make sure that the tests for the other packages still work as expected. 

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/packages/schema-blocks/README.md
+++ b/packages/schema-blocks/README.md
@@ -1,0 +1,18 @@
+# Schema blocks package
+
+## Testing
+
+### Running the tests
+You can run the tests by calling `yarn test` in the root folder of this package.
+
+Running `yarn test --coverage` runs the tests with code coverage enabled.
+
+### Adding tests.
+The tests are located in the `tests/` folder. This folder should follow the same directory structure as the `src/` folder.
+
+The tests should follow this naming convention:
+```
+{name_of_file_under_test}.test.ts
+```
+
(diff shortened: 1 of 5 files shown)


Xyfi:
CR 👍

PR #918, opened 2020-11-04
Refactor flesch reading research

FAMarfuaty:
## Summary

* [yoastseo] Make flesch reading research non-language independent.

* 

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #https://yoast.atlassian.net/browse/LINGO-525


manuelaugustin (changes requested):
CR 🚧

I think one function should be changed, see my comment.

Otherwise I just made one small change to the documentation: when the parameter is the paper, you can also say that the type is a `Paper`. It's correct that the paper is also an object, but it's a specific type of object with its own constructor that sets specific properties. So it's better to be more specific in the parameter description. (I know that this isn't done consistently in the codebase, but it's good to improve on these details when refactoring.)

manuelaugustin:
CR2 👍

PR #924, opened 2020-11-06
Refactor Hungarian Researcher

FAMarfuaty:
## Summary

* [yoastseo] Deletes all the Hungarian researches and refactors the Researcher. Also changes the way the availability of function words list is checked in `keyphraseDistribution`, `keyphraseLength` and `matchKeywordInSubheadings` researches.

* `functionWords` config is added in the Researcher and assigned to an empty array as currently Hungarian doesn't have the list. This approach is picked instead of using `.hasConfig` method because more assessments use an array rather than a boolean. Because of this, the way the availability of function words list is checked in `keyphraseDistribution`, `keyphraseLength` and `matchKeywordInSubheadings` researches is also changed. E.g. `functionWords.length === 0` or `functionWords.length > 0`.

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #https://yoast.atlassian.net/browse/LINGO-515


agnieszkaszuba:
CR: 👍 
There were some instances where hasConfig was used instead of getConfig. I changed those, and also removed hasConfig from the AbstractResearcher since it is not used anywhere anymore.

PR #933, opened 2020-11-11
Add additional languages to reading time helper

FAMarfuaty:
## Summary

* [yoastseo] Add additional languages to reading time helper.

* For languages that are not on the list, the average of the total scores is used instead.

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #https://yoast.atlassian.net/browse/LINGO-532


manuelaugustin:
CR 👍

I've just made some small changes:
- I added a citation to the paper for the reading times
- I've moved the variables that were only used within the if condition to the if condition. That's because it's not necessary to calculate them if the `if` condition doesn't apply.

PR #940, opened 2020-11-16
Refactor default researcher

FAMarfuaty:
## Summary

* [yoastseo] Deletes all the researches inside the default folder and refactors the default Researcher

* 

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #https://yoast.atlassian.net/browse/LINGO-523


manuelaugustin:
Looks great overall, I just have a question with regards to `.toLocaleLowerCase` and adding the extra researchers. As far as I can see we never use `toLocaleLowerCase` with a language argument in the code. The only exception are the language-specific `getWordForms` functions, where we specify the language within those functions. So that means the Turkish/Lithuanian/Azeri language argument is set in the researchers, but never gets used. 

So if we want to make those researchers meaningful, we'd also have to pass the language argument. Or do I not see the situation correctly?

(Also I'm not saying that that's necessarily the best way to go, because we already have the transliteration step. So it might be superfluous to also let the `toLocaleLowerCase` do part of the transliteration job. That migjht depend on the order in which you call them.)

manuelaugustin:
CR 👍

PR #943, opened 2020-11-17
Remove locale from Sentence and SentencePart objects

manuelaugustin:
## Summary

* Removes `locale` property from the `Sentence` and `SentencePart` objects used for the passive voice assessment.

### Sentence
As far as I can see, we only create `Sentence` objects in `getPassiveVoice`. However, locale isn't set there. From this I conclude that we can remove the locale from the `Sentence` object.

### SentencePart
- We create sentence parts in the language-specific sentence part constructors, where we set the locale.
- These are called in `getSentencePartsSplitOnStopwords`, where the sentence parts are exported as `splitSentence`. This in term is used in `getSentenceParts`, which is saved as a helper on the researcher. This data is used in `isPassiveSentencePart`. This function is called in `getPassiveVoice`, where it's called with the `sentencePart.getSentencePartText()` and `sentencePart.getAuxiliaries()` as parameters. `sentencePart.getlocale()` never seems to be called anywhere. From this I conclude that we can remove the locale from the `SentencePart` object.

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LINGO-535


FAMarfuaty (approved):
CR 👍 

I deleted `locale` from some files that were not deleted yet (can be seen from the last commit)

PR #944, opened 2020-11-18
Add support for templates to inner blocks

andizer:
🚨  **Note**: Test this against https://github.com/Yoast/wordpress-seo/pull/16397

## Summary

* Adds the possibility to define templates for each Innerblocks component.

* Made some changes together with @herregroen. These changes will give the possibility to have nested arrays / objects and the combination of both.

## Test instructions

* Test this on `wordpress-seo` with https://github.com/Yoast/wordpress-seo/pull/16397. Link the monorepo. In `src/schema-templates/recipe.block.php` add this to line 8:
  ```js
  template=[ [ "yoast/ingredients" ] ]
  ```
  So it will look like this:
  ```js
  {{inner-blocks allowed-blocks=[ "core/paragraph", "core/image", "yoast/ingredients" ] template=[ [ "yoast/ingredients" ] ] appender="button" appenderLabel="Add to recipe" }}
  ```
* Adds the recipe block to your post and see that the ingredients is added to it by default. This is because it defined in the template.

## Impact check

  * Post edit -> The recipe block. Has not been released yet.

## UI changes

## Quality assurance

Fixes #


packages/schema-blocks/src/functions/process.ts
```diff
@@ -39,25 +36,49 @@ function processArray( tokens: IToken[] ): unknown[] {
...
 }
 
+/**
+ * Processes a token from a list of tokens.
+ *
+ * @param tokens The remaining tokens.
+ *
+ * @returns The value of the first token.
+ */
+function processToken( tokens: IToken[] ): InstructionValue {
```
hansjovis, line 65: I would use the same method interface as `processBlockInstruction`, meaning that it would get a  `token` argument and an array of remaining `tokens`.

In my opinion it looks a bit cleaner when we consume the token in one place, instead of in multiple methods (`processToken`, `processArray` and `processObject`).

E.g.:
```js
function processToken( token: IToken, tokens: IToken[] ): InstructionValue {
	if ( token.isA( "array-open" ) ) {
		return processArray( tokens );
	}
	if ( token.isA( "object-open" ) ) {
		return processObject( tokens );
	}
	if ( token.isA( "empty-object" ) ) {
		return {};
	}
	if ( token.isA( "value" ) ) {
		return token.value as InstructionPrimitive;
	}
	throw "Invalid token found.";
}
```
Then in `processBlockInstruction` we would consume the token first, before processing it:
```js
const token = tokens.shift();
instruction.options[ key ] = processToken( token, tokens );
```
The first `tokens.shift()` in `processArray` and `processObject` should then be removed as well, since they are consumed in `processBlockInstruction` instead.

packages/schema-blocks/tests/instructions/Innerblocks.test.ts
```diff
@@ -0,0 +1,21 @@
+import "../matchMedia.mock";
+import InnerBlocks from "../../src/instructions/blocks/InnerBlocks";
+
+describe( "The Innerblocks class", () => {
+	it( "has set the options right.", () => {
+		const innerBlocks = new InnerBlocks(
+			1337,
```
hansjovis, line 7: Haha, nice 😛 .

packages/schema-blocks/tests/functions/tokenize.test.ts
```diff
@@ -0,0 +1,478 @@
+import tokenize from "../../src/functions/tokenize";
+
+const testData = [
```
hansjovis, line 3: This test data takes up a large portion of the file (over 80%). It may be better to extract it into its own (JSON) file and import it here.

packages/schema-blocks/tests/functions/tokenize.test.ts
```diff
@@ -0,0 +1,478 @@
...
+
+describe.each( testData )( "the tokenize function", ( data ) => {
+	it( `extracts ${data.description} from the instruction's definition value`, () => {
+		expect( tokenize( data.testString ) ).toEqual( expect.arrayContaining( data.expected ) );
+	} );
+} );
+
+
+describe( "the tokenize function", () => {
+	it( "extract a constant from the instruction", () => {
```
hansjovis, line 411:
```suggestion
	it( "extracts a constant from the instruction", () => {
```

packages/schema-blocks/tests/functions/tokenize.test.ts
```diff
@@ -0,0 +1,478 @@
...
+			type: "constant",
+			value: "{{ instruction }}",
+		} ];
+
+		const actual = tokenize( "{{ instruction }}" );
+
+		expect( actual ).toEqual( expect.arrayContaining( expected ) );
+	} );
+
+	it( "extract a definition from the instruction", () => {
```
hansjovis, line 426:
```suggestion
	it( "extracts a definition from the instruction", () => {
```

hansjovis (changes requested):

packages/schema-blocks/src/core/Instruction.ts
```diff
@@ -1,7 +1,11 @@
 import { RenderEditProps, RenderSaveProps } from "./blocks/BlockDefinition";
 
-export type InstructionOptions =
-	Record<string, string | boolean | number | Array<string | boolean | number> | Record<string, string | boolean | number>>;
+export type InstructionPrimitive = string | number | boolean;
+export type InstructionValue = InstructionPrimitive | InstructionObject | InstructionArray;
+export type InstructionObject = { [member: string]: InstructionValue };
```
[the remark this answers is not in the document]
andizer: I don't think so. If I use that syntax the editor tells me it will be circular.

packages/schema-blocks/src/functions/process.ts
andizer: I have adapted your suggestion partially. I tried to shift the token and passing it in processToken, but that leads to unpredictable results: the tests is hanging because of an infinite loop.

For now I keep it as it is right now, because I can't oversee the consequences of the change and it is also out of scope for this issue. Feel free to create a ticket / PR to have this code structured the way you suggested.

johannadevos:
CR on the latest commits: ✅ 

Acceptance: 🚧 

Without adding `template=[ [ "yoast/ingredients" ] ]` to `recipe.block.php`, I can add the recipe block in my editor.
Once I add this code, I can no longer add the recipe block, and I see the following in the console:

Screenshot 2020-11-20 at 15 10 48

johannadevos:
Acceptance: ✅ 
Tested with Yoast/wordpress-seo#16397

PR #963, opened 2020-12-08
P2-540 additional unit tests for p2 212

increddibelly:
## Summary

* Adds unit tests for InnerBlocks validation

* I separated some concerns into classes e.g. `BlockDefinitionRepository` 
* the interface of the `valid()` method no longer accepts SaveProps or EditProps but takes a BlockInstance instead; this removes many dependencies on the gutenberg data models across the Instructions and Definitions.
* changed the `process` method to hide implementation details and remove more dependencies.

## Test instructions

* All affected code is covered directly by new or existing unit tests; this means that Travis has done our testing for us. There is currently no way to access this code from within a WordPress installation.

## Impact check

  * 

## UI changes

## Quality assurance

Fixes #


packages/schema-blocks/src/functions/validators/innerBlocksValid.ts
```diff
@@ -78,46 +84,58 @@ function validateBlocks( requiredBlocks: RequiredBlock[], blocks: BlockInstance[
...
+	const block: InvalidBlock = {
+		name: name,
 		reason,
-	} as InvalidBlock;
+	};
+	return block;
+}
+
+/**
+ * Helper function to determine the urgency of an invalidation.
```
johannadevos, line 128: If this is a helper function, shouldn't it be placed in `innerBlocksHelper.ts`?

packages/schema-blocks/src/functions/validators/innerBlocksValid.ts
```diff
@@ -78,46 +84,58 @@ function validateBlocks( requiredBlocks: RequiredBlock[], blocks: BlockInstance[
...
 
 	// Find all blocks that have decided for themselves that they're invalid.
-	invalidBlocks.push( ...validateBlocks( requiredBlocks, innerBlocks ) );
+	// InvalidBlocks.push( ...findSelfInvalidatedBlocks( blockInstance, requiredBlocks ) );
 
 	return invalidBlocks;
 }
 
 /**
  * Helper function to create an invalid block.
```
johannadevos, line 112: If this is a helper function, shouldn't it be placed in `innerBlocksHelper.ts`?

johannadevos:
CR: ✅ with a few minor remarks.

* The changelog entry doesn't really correspond to the changes in the PR anymore. While it is not user-facing, it may be used by the plugin testers in the end, and it will end up on GitHub.
* I'm slightly surprised to see so many changes to the `yarn.lock`, is that all a consequence of bumping two packages in `package.json`?
* Two questions about helper functions.
* General remark: I did this CR by reading and clicking through the code, and I didn't notice anything that seemed strange to me. At the same time, it's hard (at least for me) to really oversee the functionality changes just by reading and clicking.
* But: Travis passes, so that means that acceptance is ✅

packages/schema-blocks/src/functions/innerBlocksHelper.ts
```diff
@@ -1,19 +1,18 @@
...
  *
- * @param blockNames     The names of the blocks you're searching for.
- * @param blockInstances The array of blocks you're searching in.
+ * @param blockInstance The block whose InnerBlocks you're searching in.
+ * @param blockNames    The names of the blocks you're searching for.
  *
  * @returns {BlockInstance[]} The block instances that have a name that occurs in blockNames.
  */
-function getInnerblocksByName( blockNames: string[], blockInstances: BlockInstance[] ): BlockInstance[] {
+function getInnerblocksByName( blockInstance: BlockInstance, blockNames: string[] ): BlockInstance[] {
```
hansjovis, line 12: Proposition: replace this with a `filterBlocksRecursively` function that takes a callback function as a second argument instead. This should make it a lot more flexible and thus re-useable.

It could look something like this:
```ts
function getInnerblocksByName( blockInstance: BlockInstance, predicate: ( blockInstance: BlockInstance ): boolean ): BlockInstance[] {
	const foundBlocks: BlockInstance[] = [];

	recurseOverBlocks( blockInstance.innerBlocks, ( block: BlockInstance ) => {
		// Checks if the current block is one of the required types.
		if ( predicate( block ) ) {
			foundBlocks.push( block );
		}
	} );

	return foundBlocks;
}
```

packages/schema-blocks/src/functions/innerBlocksHelper.ts
```diff
@@ -23,16 +22,4 @@ function getInnerblocksByName( blockNames: string[], blockInstances: BlockInstan
...
- *
- * @returns {BlockInstance[]} The block's inner blocks.
- */
-function getInnerBlocks( clientId: string ): BlockInstance[] {
-	/* istanbul ignore next */
-	return select( "core/block-editor" ).getBlock( clientId ).innerBlocks;
-}
-
-export { getInnerBlocks, getInnerblocksByName };
+export { getInnerblocksByName };
```
hansjovis, line 25: The brackets are unnecessary here 🙂.
```suggestion
export getInnerblocksByName;
```

packages/schema-blocks/src/functions/validators/innerBlocksValid.ts
```diff
@@ -30,19 +31,19 @@ function findMissingBlocks( requiredBlocks: RequiredBlock[], existingRequiredBlo
...
  * @returns {InvalidBlock[]} The names of blocks that occur more than once in the inner blocks with reason 'TooMany'.
  */
-function findRedundantBlocks( requiredBlocks: RequiredBlock[], existingRequiredBlocks: BlockInstance[] ): InvalidBlock[] {
+function findRedundantBlocks( existingRequiredBlocks: BlockInstance[], requiredBlocks: RequiredBlock[] ): InvalidBlock[] {
 	const onlyOneAllowed: string[] = [];
 	const invalidBlocks: InvalidBlock[] = [];
 
 	requiredBlocks
 		.filter( block => block.option === RequiredBlockOption.One )
 		.forEach( block => onlyOneAllowed.push( block.name ) );
```
hansjovis, line 45: The `forEach` is not necessary here. The `filter` function already contains all the `onlyOneAllowed` blocks.
```suggestion
	onlyOneAllowed = requiredBlocks
		.filter( block => block.option === RequiredBlockOption.One );
```

hansjovis (commented):
First round of feedback. Will do the second round tomorrow.

hansjovis:
I think we are confounding two different concepts within the schema blocks: the concept of a **valid** block and the concept of a **missing** block.

A **valid** block is a block that has a value (and perhaps includes other validation steps in the future). This validation would power [the green check marks behind the block names in the schema analysis in the sidebar](https://www.sketch.com/s/ab6ebb5c-a390-434c-bdf1-8e3240cadcef/a/37RLrG).

A **missing** block is an inner block that is missing from the schema block. An inner block can be either required, recommended or optional.

hansjovis:
I thought long and hard about this, and I think the inner blocks validation can be made a lot simpler:
```js
function countRecursively( innerBlocks, predicate ) {
	const counts = countBy( innerBlocks, predicate );

	return innerBlocks.reduce(
		( prev, block ) => {
			const innerBlockCount = countRecursively( block.innerBlocks, predicate );
			return mergeWith( prev, innerBlockCount,
				( prevCount, addCount ) => {
					return prevCount + addCount;
				}
			);
		},
		counts
	);
}
```
```js
function valid( blockInstance ) {
	const innerBlocks = blockInstance.innerBlocks

	// Find the number of occurrences of each inner block recursively.
	const blockCounts = countRecursively( innerBlocks, block => block.name );

	const innerBlocksValidation = this.options.innerBlocksValidation;

	// Step 1: Check if the inner blocks occur the right amount of times.
	let invalid = Object.keys( innerBlocksValidation ).some(
		blockName => {
			if ( innerBlocksValidation[ blockName ].includes( 'required' ) ) {
				return ! ( blockName in blockCounts );
			}
			if ( innerBlocksValidation[ blockName ].includes( 'unique' ) ) {
				return blockCounts[ blockName ] > 1;
			}
			return false;
		}
	);

	// If it is already considered invalid, no need to do any further checking.
	if ( invalid ) {
		return false;
	}

	// Step 2: Check if every inner block (that has a definition) is valid according to its own validation.
	invalid = innerBlocks.some(
		innerBlock => {
			const blockDefinition = getBlockDefinition( innerBlock.name );

			if ( blockDefinition ) {
				return ! blockDefinition.valid( innerBlock );
			}
		}
	)

	return ! invalid;
}
```

packages/schema-blocks/src/functions/validators/innerBlocksValid.ts
```diff
@@ -78,46 +84,59 @@ function validateBlocks( requiredBlocks: RequiredBlock[], blocks: BlockInstance[
...
  * @param reason The reason this block is invalid.
  *
  * @returns {InvalidBlock} The invalid block.
  */
-function createInvalidBlock( type: string, reason: InvalidBlockReason ): InvalidBlock {
-	return {
-		name: type,
+function createInvalidBlock( name: string, reason: InvalidBlockReason ): InvalidBlock {
+	const block: InvalidBlock = {
+		name: name,
```
andizer, line 121: You can use `name` only because both key and value are the same.
```suggestion
	name,
```

packages/schema-blocks/src/functions/validators/innerBlocksValid.ts
```diff
@@ -78,46 +84,59 @@ function validateBlocks( requiredBlocks: RequiredBlock[], blocks: BlockInstance[
...
+ *
+ * @param reason The reason to check.
+ *
+ * @returns {boolean} True if the invalidation is for an optional block (a warning), false if the invalidation is for a required block (an error).
+ */
+function isOptional( reason: InvalidBlockReason ): boolean {
+	switch ( reason ) {
+		case InvalidBlockReason.Optional: return true;
+		default: return false;
+	}
```
andizer, line 138: Please use some newlines to make this more readable
```suggestion
	switch ( reason ) {
		case InvalidBlockReason.Optional: 
			return true;
		default: 
			return false;
	}
```

andizer (changes requested):
Cr done

packages/schema-blocks/src/functions/validators/innerBlocksValid.ts
increddibelly: I split these classes on purpose : thisclass is purely about validation of innerblocks. Validation is another concern than manipulating innerblocks, which is what the innerblocksHelper does.

packages/schema-blocks/src/functions/validators/innerBlocksValid.ts
increddibelly: see  https://github.com/Yoast/javascript/pull/963#discussion_r540781410

packages/schema-blocks/src/functions/innerBlocksHelper.ts
increddibelly: that's a really nice idea.

increddibelly: done

hansjovis (approved):
Code review comments from the first round of feedback have been resolved, so I am giving my OK.

johannadevos:
Merging this PR, because CR was ✅ and no acceptance test is required.

PR #964, opened 2020-12-08
Configured JSX for the schema-blocks package.

hansjovis:
## Summary

* Adds support for JSX to the schema-blocks package.

* I disabled the ESLint rule that checks whether `React` is imported, since we use `wp.element.createElement` instead (see also the ESLint configuration in the plugin).

## Test instructions

* Replace a `createElement` function within a file in the schema-block package with the appropriate JSX syntax.
  * For example https://github.com/Yoast/javascript/blob/develop/packages/schema-blocks/src/instructions/blocks/abstract/RichTextBase.ts#L57
    with:
    ```js
    return <WordPressRichText { ...attributes } />;
    ```
* Give the file a `.tsx` filetype.
  * The `.tsx` ending makes sure that TypeScript correctly identifies the React elements in the file.
* Build the schema-blocks package and the JS files in the plugin.
* Test out the recipe block, it should still work as expected.

## Impact check

  * N/A

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/packages/schema-blocks/.eslintrc
+++ b/packages/schema-blocks/.eslintrc
@@ -77,6 +77,8 @@
                 "requireReturnType": false,
                 "requireParamType": false
             }
-        ]
+        ],
+        // We use `wp.element.createElement` instead of the `react` package directly.
+        "react/react-in-jsx-scope": 0
     }
 }
--- a/packages/schema-blocks/tsconfig.json
+++ b/packages/schema-blocks/tsconfig.json
@@ -4,6 +4,7 @@
         "noImplicitAny": true,
         "target": "es5",
         "jsx": "react",
+        "jsxFactory": "createElement",
         "allowJs": true,
         "lib": [
             "DOM",


johannadevos:
CR: ✅

johannadevos:
Acceptance: 🚧 

I made the change you suggested (`return <WordPressRichText { ...attributes } />;`).
I then ran `yarn build` in `packages/schema-blocks`. This immediately resulted in the following error:

Screenshot 2020-12-08 at 16 15 06

The error occurs irregardless of whether the file is renamed.
Because of the error, I didn't test the recipe block in the plugin yet.

hansjovis:
@johannadevos That was my mistake... 🤦 .  I forgot to add 'so it ends in `.tsx`' to the line saying you need to rename the file. I updated the line so it reads:
> Rename a file that uses createElement to create a React element in the schema-blocks package so it ends in `.tsx`.

The `.tsx` ending communicates to TypeScript that the file contains JSX.

johannadevos:
Acceptance: ✅ 
- [x] Building succeeds without problems.
- [x] The recipe block functions as normal.

PR #979, opened 2020-12-16
P2-505 Adds required blocks to sidebar

andizer:
## Summary

* Introduces a panel in the sidebar containing the required blocks with the possibility to add a block that isn't added yet.

* At the moment the block doesn't contain the checkmark icon for the blocks that are added already. The reason for this is that a decision should be made on which icon to use for that.

## Test instructions

enable the schema blocks feature by adding the following line to loader.php below the use statements:
`\define( 'YOAST_SEO_SCHEMA_BLOCKS', true );`
* start a new post
* show the sidebar
* add a recipe block
* see the "paragraph" block in the sidebar; press add
* see there's a new paragraph in the post; the sidebar says OK
* remove the paragraph block; the sidebar allows you to add again.

the required blocks are goverened by recipe.block.php. you can play around with this a little but it's quite technical:
```{{inner-blocks required-blocks=[ { "name": "core/paragraph", "option": "One" } ] ```
```{{inner-blocks required-blocks=[ { "name": "core/paragraph", "option": "One" }, { "name": "core/image", "option": "Multiple" } ] ```

* Link the monorepo, do a `yarn build` in the schema templates package and do a `grunt build` in wordpress-seo

## Impact check

  * the feature is still disabled and should not have any impact on existing code.

## UI changes

## Quality assurance

Fixes #

### CODE REVIEW

File: packages/schema-blocks/src/blocks/RequiredBlocks.tsx [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+		};
+
+		requiredBlockItems.push(
+			(
+				<li key={ index } className="yoast-block-suggestion">
+					{ blockType.title }
+					<button className="yoast-block-suggestion-button" onClick={ addBlockClick }>Add</button>
+				</li>
+			),
+		);
```
hansjovis, line 66: Same here: I would add this to a separate component.
```js
function BlockSuggestion( { blockClientId, blockTitle, blockName } ): ReactElement {
	/**
	 * Onclick handler for the remove block.
	 */
	const addBlockClick = () => {
		const blockToAdd = createBlock( blockName );
		insertBlockToInnerBlock( blockToAdd, blockClientId );
	};

	return (
		<li className="yoast-block-suggestion">
			{ blockTitle }
			<button className="yoast-block-suggestion-button" onClick={ addBlockClick }>Add</button>
		</li>
	);
}
```
And then:
```js
requiredBlockItems.push( 
	<BlockSuggestion 
		key={ index } 
		blockClientId={ block.clientId } 
		blockTitle={ blockType.title } 
		blockName={ requiredBlockName }
	/>
);
```

packages/schema-blocks/src/blocks/RequiredBlocks.tsx
```diff
@@ -0,0 +1,81 @@
...
+	} );
+
+	const requiredBlockItems: ReactElement[] = [];
+
+	requiredBlockNames.forEach( ( requiredBlockName: string, index: number ) => {
+		const blockType = getBlockType( requiredBlockName );
+
+		if ( typeof blockType === "undefined" ) {
+			return;
+		}
```
hansjovis, line 36: I am pondering whether we can remove this check...

Removing this would mean that we can transform this `forEach` + `push` loop into a `map`. Since a block is either missing or not. 

Though it can be worthwhile to do some defensive programming 🤔. 

Or is this possibly a case where we can apply exception handling? Since we expect a block type to be registered for a required block name. Perhaps failing silently, as we do here is not what we want, since we may want to indicate that something went wrong with an error message in the console.

Just my thoughts 😛. What do you think @andizer?

packages/schema-blocks/tests/blocks/RequiredBlocks.test.ts
```diff
@@ -0,0 +1,109 @@
+import { BlockInstance, createBlock } from "@wordpress/blocks";
+
+import renderer from "react-test-renderer";
```
hansjovis, line 3: I recently found out why TypeScript gives an error on this import in PHPStorm and how to solve this:
```suggestion
import renderer as * from "react-test-renderer";
```
There is indeed no default export. Importing it like this solves it.

I solved this in my other PR, you may want to do the same here 🙂.

packages/schema-blocks/src/functions/BlockHelper.ts
```diff
@@ -34,4 +34,15 @@ function restoreBlock( clientId: string, removedBlock: string, removedAttributes
...
 
-export { getBlockByClientId, removeBlock, restoreBlock };
+/**
+ * Finds the block type in the list of registered blocks in the block editor.
+ *
+ * @param {string} blockName The block to search for.
+ *
+ * @return {BlockInstruction} The found block.
+ */
+function getBlockType( blockName: string ) {
```
hansjovis, line 44: I am missing a TypeScript return type:
```suggestion
function getBlockType( blockName: string ): Block | undefined {
```

hansjovis: Do not forget to import the type definition of `Block` at the top:
```js
import { Block, BlockInstance, createBlock } from "@wordpress/blocks";
```

packages/schema-blocks/src/functions/innerBlocksHelper.ts
```diff
@@ -33,4 +34,15 @@ function filterBlocksRecursively( blockInstance: BlockInstance, predicate: ( blo
...
 
-export { filterBlocksRecursively, getInnerblocksByName };
+/**
+ * Inserts a block to the inner block.
+ *
+ * @param {BlockInstance} block    The block to insert.
+ * @param {string}        clientId Id of the element to insert the block to.
+ * @param {number}        index    The location of the block.
+ */
+function insertBlockToInnerBlock( block: BlockInstance, clientId: string, index?: number ) {
```
hansjovis, line 44: This should have a `void` return type.
```suggestion
function insertBlockToInnerBlock( block: BlockInstance, clientId: string, index?: number ): void {
```

hansjovis (changes requested):

packages/schema-blocks/tests/blocks/RequiredBlocks.test.ts
hansjovis: I made a mistake, my bad 😬 . But we figured it out.

It should have been:
```js
import * as renderer from "react-test-renderer";
```

packages/schema-blocks/src/blocks/RequiredBlocks.tsx
```diff
@@ -17,64 +69,42 @@ import { PanelBody } from "@wordpress/components";
...
-		return presentBlock.name;
-	} );
-
-	const requiredBlockItems: ReactElement[] = [];
-
-	requiredBlockNames.forEach( ( requiredBlockName: string, index: number ) => {
-		const blockType = getBlockType( requiredBlockName );
+	const requiredBlockNames = requiredBlocks
+		.filter( requiredBlock => {
+			const blockType = getBlockType( requiredBlock.name );
```
hansjovis, line 74: Without this intermediate variable it can be even shorter:
```js
.filter( requiredBlock => typeof getBlockType( requiredBlock.name ) !== "undefined" )
```

packages/schema-blocks/tests/blocks/RequiredBlocks.test.ts
```diff
@@ -1,11 +1,11 @@
...
 
-import renderer from "react-test-renderer";
-import { shallow } from "enzyme";
+import * as renderer from "react-test-renderer";
+import { mount } from "enzyme";
 
 import { RequiredBlock } from "../../src/instructions/blocks/dto";
 import RequiredBlocks from "../../src/blocks/RequiredBlocks";
-import { insertBlockToInnerBlock } from "../../src/functions/innerBlocksHelper";
+import { insertBlocks } from "../../src/functions/innerBlocksHelper";
```
hansjovis, line 8:
```suggestion
import { insertBlock } from "../../src/functions/innerBlocksHelper";
```
Plus, there are some other references in this file that still reference `insertBlocks`.
Also, the mock of `insertBlockToInnerBlock` should be changed to mock `insertBlock` instead.

hansjovis (commented):
Two small things. I am going to fix them myself real quick.

PR #995, opened 2021-01-05
P2-217 Create warning subscriber

hansjovis:
## Summary

* Shows a warning whenever a required or recommended block is removed from a schema block, giving the user the option to restore the block again.

* The `parent: []` method to remove the ability for users to add the warning block manually did not work as expected. It also prevented adding the warning block programmatically.
	* We found a better way by [telling WordPress that inserting via the block inserter is not supported](https://developer.wordpress.org/block-editor/developers/block-api/block-supports/#inserter).
* `restoreBlock` and the warning block now get the entire block instance, instead of the data to restore the block separately.
	* Since we also needed to restore the innerBlocks, this was the most efficient way.
* Added a `mapBlocksRecursively` helper function that maps a callback recursively over innerBlocks and returns the results in a flat array. 

## Test instructions

* Add these options to the `innerBlock` instruction in the `recipe.block.php` template in the WordPress SEO plugin:
  ```
  required-blocks=[ { "name": "yoast/steps", "option": "1", "warning": "<?php esc_html__( "You've just removed the ‘Ingredients’ block, but this is a required block for Schema output. Without this block no Schema will be generated. Do you want this?" ); ?>" } ] recommended-blocks=[ { "name": "yoast/ingredients" } ]
  ```
  * This tells the schema-blocks package that the `yoast/steps` blocks is required, the `yoast/ingredients` block is recommended and to show a specific warning message when the ingredients block is removed.
* Create a new post.
* Add a recipe schema block to the post.
* Add an ingredients block to the recipe.
	* Add some ingredients.
* Add a steps block to the recipe.
	* Add some steps.

### Removing a recommended block.  
* Remove the ingredients block.
* You should see a yellow warning message indicating that this is a recommended block.
* Click on 'No, please undo this'.
* The ingredients block should be restored in its original place, including the individual ingredients.
* Remove the ingredients block again.
* You should see the warning message again.
* Click on 'Yes'.
* The warning block should disappear.

### Removing a required block.
* Remove the steps block.
* You should see a red warning message indicating that this is a required block.
* Click on 'No, please undo this'.
* The steps block should be restored in its original place, including the individual steps.
* Remove the steps block again.
* You should see the warning message again.
* Click on 'Yes'.
* The warning message should disappear.

## Impact check

  * It only affects the unreleased schema block functionality.

## UI changes

## Quality assurance

Fixes #


Xyfi:
CR 👍

packages/schema-blocks/src/blocks/warning-block/configuration.ts
```diff
@@ -32,9 +29,11 @@ export const WarningBlock: BlockConfiguration = {
 	},
 
 	/**
-	 * Makes sure users cannot select this block.
+	 * Make sure that users cannot add this block manually.
 	 */
-	parent: [],
+	supports: {
```
increddibelly, line 34: great!

Xyfi:
Acceptance 🚧 

I would argue that the warning message should be part of the required-blocks attribute (also for the recommended blocks attribute), like this: `required-blocks=[{name:"yoast/recipe", option:"1", warning: "..."}]`.

We could also opt for a default message. Because "You've just removed the '%s' block, but this is a required block for Schema output. Without this block no Schema will be generated. Do you want this?" seems like a very generic message for required blocks.

packages/schema-blocks/src/functions/gutenberg/watchers/warningWatcher.ts
```diff
@@ -0,0 +1,216 @@
...
+ * Returns the default warning message for a required or recommended block.
+ *
+ * @param blockTitle The title of the removed block (e.g. 'Recipe').
+ * @param warningType The type of warning to show.
+ *
+ * @returns The default warning message.
+ */
+function getDefaultWarningMessage( blockTitle: string, warningType: WarningType ): string {
+	switch ( warningType ) {
+		case WarningType.BLOCK_REQUIRED: {
```
increddibelly, line 49: <3

Xyfi:
CR 👍 
Acceptance 👍

PR #997, opened 2021-01-05
Create ImageSelect component

JoseeWouters:
## Summary

* Adds an ImageSelect component with an image preview or url preview

*

## Test instructions

* Run `yarn start` in `apps/components`
* Go to image select
* When clicked on the image preview or "select image" button, a placeholder image should appear
* Make sure you can also remove the image by clicking "remove image" button
* In ImageSelectWrapper.js change `hasPreview` to `false` and you should now see a disabled input field
* With an image selected, this input field should show the url of the image

## Impact check

  *

## UI changes

## Quality assurance

Fixes P1-187


apps/components/ImageSelectWrapper.js
```diff
@@ -0,0 +1,59 @@
...
+	 */
+	render() {
+		return (
+			<div className="yoast">
+				<h2>ImageSelect</h2>
+				<ImageSelect
+					imageUrlInputId="test-image"
+					imageUrl={ this.state.imageUrl }
+					imageSelected={ !! this.state.imageUrl }
+					hasPreview={ false }
```
igorschoester, line 47: Maybe add another ImageSelect that has `hasPreview` true? Or a toggle so you can switch between?

packages/components/src/image-select/ImageSelect.js
```diff
@@ -0,0 +1,65 @@
...
+
+export default ImageSelect;
+
+ImageSelect.propTypes = {
+	imageUrl: PropTypes.string,
+	imageAltText: PropTypes.string,
+	imageSelected: PropTypes.bool,
+	hasPreview: PropTypes.bool.isRequired,
+	label: PropTypes.string.isRequired,
+	onClick: () => {},
```
igorschoester, line 57:
```suggestion
	onClick: PropTypes.func,
```

packages/components/src/image-select/ImageSelect.js
```diff
@@ -0,0 +1,65 @@
...
+ * @returns {React.Component} The ImageSelect.
+ */
+class ImageSelect extends React.Component {
+	/**
+	 * Returns the rendered HTML.
+	 *
+	 * @returns {React.Component} The ImageSelect.
+	 */
+	render() {
+		const imageSelectButtonsProps = this.props;
```
igorschoester, line 19: I think it is better to specify which props are for the ImageSelectButtons exactly. It makes it transparent and it will prevent unintentional extra props like `ImageAltText` from being passed on.
Also the IDs and `onRemoveImageClick` are missing from the ImageSelect props.

packages/components/src/image-select/ImageSelect.js
```diff
@@ -0,0 +1,65 @@
...
+import PropTypes from "prop-types";
+import FieldGroup from "../field-group/FieldGroup";
+import TextInput from "../inputs/TextInput";
+
+/**
+ * Renders ImageSelect
+ *
+ * @returns {React.Component} The ImageSelect.
+ */
+class ImageSelect extends React.Component {
```
igorschoester, line 12: Because this component has no complex features (even state doesn't count anymore with hooks 😄 ), it would be nice if this was a function instead of a class.

igorschoester (changes requested):
CR 🏗️ 

Besides the code comments: the tests are missing 😅

packages/components/src/image-select/ImageSelect.js
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Returns the rendered HTML.
+	 *
+	 * @returns {React.Component} The ImageSelect.
+	 */
+	render() {
+		const imageSelectButtonsProps = this.props;
+		const imageClassName = this.props.imageSelected
```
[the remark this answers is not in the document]
JoseeWouters: imageSelected prop exists already and eslint doesn't allow a negated condition. Therefore keeping it like this.

apps/components/ImageSelectWrapper.js
JoseeWouters: Made a toggle, great idea!

marijnyoast (approved):
CR 👍🏻

maartenleenders:
I fixed the snapshots, but they are caused by the inclusion of the `TextInput`, which I believe is not necessary for this PR:

@JoseeWouters Is there a decision anywhere about showing the disabled input URL when the preview is hidden?

If not, I believe you may have misunderstood my intention when I said "make the preview conditional". 

My intention was not to bring back the disabled URL input when the preview is hidden, but to either have "only buttons", or "buttons and preview". I believe the agreement was that everywhere where there is an image select + input field now (e.g. settings, social previews in free) we will eventually have that preview variant you made.

In the case of premium, social previews have "their own preview", so we only need the `ImageSelectButtons`.

So perhaps what we actually want is to export the `ImageSelectButtons` (as you have), and a `ImageSelect` that always has a preview. No booleans required, and no chance of accidentally putting "hidden premium code" in free, which is a faux-pas.

I'd be happy to help out with implementing that.

maartenleenders:
Moved to implement the new `ImageSelect` in the Social Previews here: https://github.com/Yoast/javascript/tree/use-new-image-select-in-social-preview-form  but needs some more work later.

maartenleenders:
Acc ✅

PR #1000, opened 2021-01-06
Fix app/components

maartenleenders:
## Summary

The components app was broken by `react-hot-loader`. Removing the `AppContainer` fixed that.

Also React Hooks were broken in the components app. They are especially useful there, because they allow the creation of very shallow wrapper components for our "stateless" components.

* Removes the react-hot-loader dependency from the components app.
* Sets react and react-dom as a peer dependency in yoast-components, configuration-wizard, and schema-block packages.

* I have removed `react-hot-loader` because it could not deal with `react hooks`. I have found a solution here https://github.com/gaearon/react-hot-loader/issues/1227 which requires another `react-dom` package and some code refactoring. Instead, I quickly removed the hot-loader to see whether that worked, and it seems to me that we then still have a hot-loading functionality via webpack. So I deemed react-hot-loader unnecessary. But there might be something I have missed, so this needs some extra eyes.
* React Hooks were still not available in the components app, and set `react` as a peer-dependency (almost) everywhere. We had multiple `react` versions which seemed to violate the criteria set out here: https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react

## Test instructions

* `yarn install` in the monorepo root, and `yarn start` in `apps/components`. Go to `localhost:3333`
* Test `useState`: I have added a buttonwrapper in the components `app`: there is a button called "test usestate" in the `Reactified Components` section, and it uses `useState`. Clicking it should update the count. ⚠️ You could undo the changes I made to the ReactifiedComponentsWrapper before merging, if you'd like.
* Test "hot reloading": Change some things in the components app (like, add a simple `button`, change some button "variants". Verify that the components app reloads to reflect your changes.

## Impact check

  *

## UI changes

## Quality assurance

### DIFF
--- a/apps/components/AppWrapper.js
+++ b/apps/components/AppWrapper.js
@@ -1,7 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Provider } from "react-redux";
-import { AppContainer } from "react-hot-loader";
 
 import DevTools from "./utils/DevTools";
 import configureStore from "./configureStore";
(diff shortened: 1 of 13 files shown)


marijnyoast:
CR 👍🏻 

Let's see whether or not we'll miss `react-hot-loader` in the future.

marijnyoast:
Acceptance 👍🏻 

Having the ability to test useState seems very functional to me. So keeping the button as well.

marijnyoast (approved):

PR #1005, opened 2021-01-07
No longer make Lerna ignore changes made to the schema-blocks package

johannadevos:
## Context
* We need to release the `schema-blocks` package because it now contains a CSS file that is imported in the plugin.
* We could do the release manually, but there no longer is any reason why we should prevent the `schema-blocks` package from being released.
* Undoing the exclusion of this package in Lerna means the least overhead for our team.

## Summary

* No longer excludes the `schema-blocks` package from being released by Lerna.

* This PR reverts the changes made in https://github.com/Yoast/javascript/commit/864c230c30310894ef0a33712b55a1e2de45255a.

## Test instructions

* When @igorschoester will have released the monorepo on Monday 11 January, we should double-check whether the `schema-blocks` package will have been released.
* I don't think there is anything we can test right now.

## Impact check

  * N/A

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/P2-530

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,8 +2,5 @@
   "npmClient": "yarn",
   "useWorkspaces": true,
   "version": "independent",
-  "includeMergedTags": true,
-  "ignoreChanges": [
-    "packages/schema-blocks/**"
-  ]
+  "includeMergedTags": true
 }


increddibelly (approved):

increddibelly (approved):

increddibelly:
CR OK

PR #1006, opened 2021-01-07
Create DurationInput component

maartenleenders:
## Summary

* Adds a DurationInput component, which can handle the input of a duration in hours, minutes, and seconds.

* Cherry-picked https://github.com/Yoast/javascript/commit/6dc7613b484c9d6a116924e1a7e38562ee00fb1d because I was rummaging in the `ReactifiedComponentsWrapper` which has been split in that commit. Then decided I was overcomplicating the PR and reverted the cherry-pick. So that's what that mess is all about.
* I really wanted to use React Hooks, which is fine in our plugin, but was somehow broken in the components app. That's why, to test this PR, you will need to merge (*without committing*) another PR where I "fix" the components app: https://github.com/Yoast/javascript/pull/1000.
* Deviates from design slightly, checked for approval from UX: https://yoast.slack.com/archives/C01FYN1QGB1/p1610026441033700 

## Test instructions

First you'll need to get the components app working:
* `git fetch`
* `git merge fix-component-app-react-hooks-by-removing-hot-loader  --no-commit`
* This will give you a merge commit, but you only have to `accept incoming change` in the ReactifiedComponentsWrapper.
* In the root of the monorepo, run `yarn install --check-files` for good measure.
* Go to `apps/components` and run `yarn start`.
* Go to localhost:3333 and checkout the `Inputs` section on the `Reactified Components` page. There is a `Duration` input there. See that it works as expected:

• No negative durations.
• No minutes and seconds not higher than 59
• Keyboard input only accepts numbers. Arrow buttons should work.

⚠️ unfortunately the fact that we use `useState` here does make the merging of #1000 a requirement for this issue.

## Impact check

  *

## UI changes

## Quality assurance

Fixes P1-292


maartenleenders:
Test failures with the follow error:
```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
```
will be fixed by #1000. You can verify that by running `yarn test` locally in the setup described in the test description (`merge --no-commit` and `yarn install --check-files`)

marijnyoast (approved):
Merged `develop` into this branch to fix `useSelect` for Travis. Renamed classnames to BEM.

CR 👍🏻

marijnyoast:
Acceptance 👍🏻

Love the fact that you can't manually input a 'forbidden' value. Better then plain HTML due to Lodash Clamp 😎

PR #1018, opened 2021-01-18
Add specs for Polish helpers, values and Researcher

FAMarfuaty:
## Summary

* [yoastseo] Adds/moves Polish languageProcessing specs. Also removes some statements that are always true or never true from the Polish stemmer and fixes the structure of some config files.

* 

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LINGO-586


agnieszkaszuba:
CR: 👍 
I changed the exports of the lists so that they are exported as arrays. I also changed how the output of getConfig is being tested in the ResearcherSpec to compare it with the output of an import. And for the getStemmerSpec example of a word stemmed with the base stemmer, I changed it to _rośliny_ since _koty_ is currently also not stemmed with the 'real' stemmer so the result is always the same.

PR #1032, opened 2021-01-27
Refactor sentence length and Flesch reading ease assessments to get scores config from the researcher

agnieszkaszuba:
## Summary

* [yoastseo] Refactors sentence length and Flesch reading ease assessments so that they get config from the researcher.

* Before, the language-specific or default config for the Flesch reading and sentence length assessments was retrieved in the contentAssessor using locale. This was changed so that instead, the default config is defined inside the assessment and any language-specific config is retrieved using the getConfig method.
* Since the assessments now use the .getConfig method on the researcher, a `config` parameter was added to the `buildMockResearcher` function inside `spec/specHelpers/factory` so that the specs for these assessments can use the mock researcher.

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #


manuelaugustin (changes requested):
CR 🚧

There are a fwe points that still need consideration:

- The config now is saved no the language-specific researchers. I think pragmatically speaking it makes sense, because that's the one place where we save language-specific config now. Conceptually, however, it blurs the line between research functionality and scoring. That's because with this approach, we now use the researcher config, which is intended to be used for researches, to make scoring decisions. Given the pragmatic considerations, I think it's still a viable approach. But it would be good to check this with an architect.
- I think it still makes sense to let the sentence length and the Flesch reading ease assessments be based on the `Assessment` class. At the moment there's not much going on in this class, but that might change in the future. So I'd rather keep that consistent across assessments.
- Previously we had specific config for some languages (IT, RU, PL, ES, CA, PT) and for other languages we had a default config. Now we have config for all languages, which means we makes copies of the default config for languages that have no specific config. I think it makes sense to keep a default config on the assessment, and only overwrite this with the specific config if there is one available.
- I think it's a good idea to make an issue specifically for the specs that we're skipping now. That way we can make sure that we won't forget them.

agnieszkaszuba:
> CR 🚧
> 
> There are a fwe points that still need consideration:
> 
> * The config now is saved no the language-specific researchers. I think pragmatically speaking it makes sense, because that's the one place where we save language-specific config now. Conceptually, however, it blurs the line between research functionality and scoring. That's because with this approach, we now use the researcher config, which is intended to be used for researches, to make scoring decisions. Given the pragmatic considerations, I think it's still a viable approach. But it would be good to check this with an architect.

I asked Herre and he agrees with the approach.

> * I think it still makes sense to let the sentence length and the Flesch reading ease assessments be based on the `Assessment` class. At the moment there's not much going on in this class, but that might change in the future. So I'd rather keep that consistent across assessments.

None of the other readability assessments that didn't have a config parameter were based on the class, so I left it this way.

> * Previously we had specific config for some languages (IT, RU, PL, ES, CA, PT) and for other languages we had a default config. Now we have config for all languages, which means we makes copies of the default config for languages that have no specific config. I think it makes sense to keep a default config on the assessment, and only overwrite this with the specific config if there is one available.

Done!

> * I think it's a good idea to make an issue specifically for the specs that we're skipping now. That way we can make sure that we won't forget them.

The specs are complete now.

manuelaugustin:
CR 👍

I've just made some small changes in the assessment (purely cosmetic, not functional) to make it a bit clearer at the first glance which one is the default config and which one is the language-specific one.

PR #1050, opened 2021-02-08
Added missing React key attributes.

hansjovis:
## Summary

* [schema-blocks] Fixes a bug where an error was shown in the browser's console when a job posting block was added to the post.

*

## Test instructions

* Create a new post.
* Add a job posting block.
* Check your browser's console, it should not have errors like these:
  ```
  Warning: Each child in a list should have a unique "key" prop.
  ```

## Impact check

  * N/A: behind a feature flag.

## UI changes

## Quality assurance

Fixes #


packages/schema-blocks/src/instructions/blocks/VariationPicker.tsx
```diff
@@ -107,7 +107,7 @@ class VariationPicker extends BlockInstruction {
 		};
 
 		return (
-			<div { ...blockProps }>
+			<div key="variation-picker" { ...blockProps }>
```
andizer, line 110: Doesn't this cause troubles when you have two blocks with a variationpicker?

andizer (approved):
I've got one question. If that is answered you can move this to needs-acceptance :)

packages/schema-blocks/src/instructions/blocks/VariationPicker.tsx
hansjovis: React uses the key to distinguish between sibling components. [Thus the variation picker's component must only be unique among its siblings.](https://reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings)

Since the variation picker is enclosed within a block (which wraps it in other components) the chance of conflicting IDs is pretty minimal. I added an index just in case.

andizer (approved):
CR 👍

PR #1052, opened 2021-02-09
P2-560 Replace console.logs with logger

increddibelly:
## Summary

* [@yoast/schema-blocks] Hides debug logging behind an abstraction.

* hides all console.logs behind an abstraction. Given this is a new package, we expect logging will help debugging in the next few weeks. This way we can enable logging on our local systems and hide it on production with a one line change in WPSEO `js/src/schema-blocks.js`

## Test instructions

* use the [branch in wpseo](https://github.com/Yoast/wordpress-seo/pull/16648) to initialize logging

## Impact check

  *

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/packages/schema-blocks/src/functions/logger.ts
+++ b/packages/schema-blocks/src/functions/logger.ts
@@ -62,11 +62,11 @@ export class ConsoleLogger implements Logger {
      * @param level The severity of the message.
      * @param input The data to log.
      */
-	protected writeLog( level: LogLevel, input: unknown[] ): void {
+	protected writeLog( level: LogLevel, ...input: unknown[] ): void {
 		if ( this.level >= level && level !== LogLevel.NONE ) {
 			// This should be the only remaining console.log in the repository.
 			// eslint-disable-next-line no-console
-			console.log( LogLevel[ level ], ":", input );
+			console.log( LogLevel[ level ] + ":", ...input );
 		}
 	}
 
--- a/packages/schema-blocks/src/functions/logger.ts
+++ b/packages/schema-blocks/src/functions/logger.ts
@@ -1,6 +1,6 @@
 
 
-enum LogLevel {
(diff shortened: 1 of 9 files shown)


packages/schema-blocks/src/functions/logger.ts
```diff
@@ -0,0 +1,116 @@
...
+     * @param level The severity of the message.
+     * @param input The data to log.
+     */
+	protected writeLog( level: LogLevel, input: unknown[] ): void {
+		if ( this.level >= level && level !== LogLevel.NONE ) {
+			// This should be the only remaining console.log in the repository.
+			// eslint-disable-next-line no-console
+			console.log( LogLevel[ level ], ":", input );
+		}
+	}
```
hansjovis, line 71: I would implement it like this:
```suggestion
	protected writeLog( level: LogLevel, ...input: unknown[] ): void {
		if ( this.level >= level && level !== LogLevel.NONE ) {
			// This should be the only remaining console.log in the repository.
			// eslint-disable-next-line no-console
			console.log( LogLevel[ level ] + ":", ...input );
		}
	}
```

It would mean the difference between
```
DEBUG : (3) ["some", "data", {…}, "stuff"]
```
and
```
DEBUG: some data {} stuff
```
The object is still inspectable, so not stringified. The same applies for arrays.

hansjovis (changes requested):
One comment about the logger.

Also: Andy has a point. Maybe we do need a way to set the flag via PHP.

We can use the same strategy as we use for the analysis web worker. That is to set a flag in the form of a constant (e.g. `YOAST_DEBUG_SCHEMA_BLOCKS`) and inject it into the JavaScript with `$asset_manager->localize_script`.

See also https://github.com/Yoast/wordpress-seo/blob/trunk/admin/taxonomy/class-taxonomy.php#L176

PR #1053, opened 2021-02-09
added a 250ms debounce to the gutenberg subscriber

increddibelly:
## Summary

* [@yoast/schema-blocks] Adds a processing delay to the Gutenberg subscriber

* we are notified of every change in the editor. This causes a lot of processing for schema generation, even though little is changed. With the debounce function, we only respond at most a few times every second, and we'll still be up to date.

## Test instructions

* open developer tools; watch for the line "generating schema output!" 
* start editing / adding blocks. 

without this PR your dev tools will be absolutely flooded with logs, a clear indication of the amount of work that's being done in the background.

with this PR, you should see significantly less logs, at most one or two "groups" of logs every second. 
Note that schema is still being generated.

## Impact check

  *

## UI changes

## Quality assurance

Fixes #


johannadevos (approved):

johannadevos:
Acceptance: ✅ 

It was a bit hard to spot a difference between this branch and `develop`, but I when I changed the value from `250` to `2500` I could clearly see the delay. So I'm concluding that it works at `250` also.

PR #1063, opened 2021-02-15
P2 671 analysis for job posting not always shown

increddibelly:
## Summary

Fixes missing analysis for job posting.

* [@yoast/schema-blocks] Improves the sidebar for Yoast structured data blocks.

* the generic Block will now check if any of the current block's parents are yoast/job-posting or yoast/recipe. if so, show that parent's sidebar instead of the current block's sidebar.
* known issue: only the analysis is shown; we need to pass the parent block's *options* too if we want further details in the sidebar.

## Test instructions

* add a job block
* navigate up/down with cursor keys until the entire job posting is selected. notice the analysis in the sidebar.
* navigate up/down to any of the job posting's innerblocks.
* notice the analysis is still in the sidebar, *but the other sidebar fields are now missing.*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #


packages/schema-blocks/src/instructions/blocks/Block.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+	/**
+	 * Renders the sidebar.
+	 *
+	 * @param props The props.
+	 *
+	 * @returns The sidebar element to render.
+	 */
+	sidebar( props: RenderEditProps ): ReactElement {
+		const rootId = this.getParentBlockOfType( props.clientId, [ "yoast/recipe", "yoast/job-posting" ] );
```
andizer, line 40: I understand why this is needed, but now the coupling between the monorepo and wordpress-seo is a bit harder. I think we should check if this is something to fix after the release.

andizer (commented):
One point of improvement, but we can fix that later :)

andizer:
Acceptance 👍

andizer:
Nieuw_bericht_toevoegen_‹_Basic_—_WordPress

PR #1078, opened 2021-02-24
Add checkbox component

igorschoester:
## Summary

* [@yoast/components] Updates the `Checkbox` component to the newer style. Now wrapped in a `FieldGroup`.
* [@yoast/components] Adds a maximum width to the `Alert` component.

* Removing the old style checkbox (I don't see it being used), the props/usage remains the same 
* Using the field group as wrapper. This is suboptimal because: 
   * the BEM `__checkbox` should be a child of the field group
   * the field group is really more a fieldset, but then this component can only be singular
* Using `useCallback` from `React` here, but in the plugin that would've been from `@wordpress/element`. Should look into this sometime. 

## Test instructions

* Tested with News PR: https://github.com/Yoast/wpseo-news/pull/678

## Impact check

  *

## UI changes

## Quality assurance

Partially fixes https://yoast.atlassian.net/browse/P1-301


packages/components/src/checkbox/Checkbox.js
```diff
@@ -0,0 +1,39 @@
+import PropTypes from "prop-types";
+import React, { useCallback } from "react";
+import { FieldGroup } from "../field-group";
+import "./checkbox.css";
+
+/**
+ * Represents the input type checkbox with a label. Optionally a FieldGroup.
```
JoseeWouters, line 7: Please remove `optionally a fieldgroup` or make it optional.

JoseeWouters (changes requested):
Please create a technical debt issue for the use of `useCallback` and one minor change for a comment.

JoseeWouters (approved):
CR 👍

marijnyoast:
Acceptance 👍🏻

PR #1108, opened 2021-03-15
Add specs/adapt code to make Czech stemmer coverage 100 %

agnieszkaszuba:
## Summary

* [yoastseo] Adds missing specs to the Czech stemmer and changes code that is redundant or a mistake. Also remakes the golden standard list and shortens it so that it is easier to remake/adapt in the future.
* [Yoast SEO Free] Improves Czech word form recognition of a specific subset of words (words that end on ště/šti/ští + suffix).

* The check for augmentative suffix -ak was deleted. The suffix -ak is already removed in the previous stemming step (diminutive suffixes). So the only case in which -ak would be removed in the augmentative suffix step would be if it had another diminutive suffix removed in the previous step. But in those cases, if the word ends in -ak it cannot be an augmentative suffix, as you wouldn't have an augmentative suffix followed by a diminutive suffix.

## Test instructions

*

## Impact check

  *

## UI changes

## Quality assurance

Fixes #


iolandasequino:
CR and Acceptance: 👍

PR #1119, opened 2021-03-25
P2 694 create schema blocks redux store

increddibelly:
## Summary

* [@yoast/schema-blocks] creates a redux store for schema blocks .

* moved reducer, actions and selectors from free ( see https://github.com/Yoast/wordpress-seo/pull/16825 ) to the schema blocks monorepo package.
* added Typescript type hints.
* organized the module structure.

## Test instructions

* add any yoast block; this will trigger validation, which should be stored in redux.
* use the redux DevTools to analyze the yoast/schema-blocks store, the validation results should be there.
image
* there should not be any console errors related to the actions e.g. `cannot find property getSchemaBlocksValidationResults of undefined`

## Impact check

  *

## UI changes

## Quality assurance

Fixes #


packages/schema-blocks/src/functions/redux/SchemaBlocksAction.ts
```diff
@@ -0,0 +1,13 @@
+import { BlockValidationResult } from "../../core/validation";
+
+const PREFIX = "WPSEO_";
+
+export const SchemaBlocksStoreActions = {
+	ADD_BLOCK_VALIDATION: `${ PREFIX }ADD_BLOCK_VALIDATION`,
+	RESET_BLOCK_VALIDATIONS: `${ PREFIX }CLEAR_BLOCK_VALIDATIONS`,
+};
+
+export type SchemaBlocksStoreAction = {
```
Xyfi, line 10: Every action in a reducer only jas one common denominator: "type". I'd recommend creating types for every action individually and define them in the `actions/index.ts` file, to prevent having to create a lot of extra imports and having to keep track where the types live.

packages/schema-blocks/src/functions/redux/actions/schemaBlocks.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+/**
+ * Commands the store to reset the block validation store to initial state.
+ * @returns {SchemaBlocksStoreAction} The command to set the initial state for the block validation store.
+ */
+export function resetBlockValidation(): SchemaBlocksStoreAction {
+	return {
+		type: SchemaBlocksStoreActions.RESET_BLOCK_VALIDATIONS,
+		validation: null,
```
Xyfi, line 25: This is an unused value.

packages/schema-blocks/src/functions/redux/reducer/schemaBlocks.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * A reducer for the Schema blocks.
+ *
+ * @param {SchemaBlocksState} state  The current state of the object.
+ * @param {SetBlockValidation} action The received action.
+ *
+ * @returns {Object} The state.
+ */
+export function schemaBlocksReducer( state: SchemaBlocksState = SchemaBlocksDefaultState, action: SchemaBlocksStoreAction ): SchemaBlocksState {
```
Xyfi, line 12: The type of action should be `AnyAction`, which can be imported from redux.

packages/schema-blocks/src/functions/redux/SchemaBlocksState.ts
```diff
@@ -0,0 +1,12 @@
+import { BlockValidationResult } from "../../core/validation";
+
+
+export const YOAST_SCHEMA_STORE_NAME = "yoast-seo/schema-blocks";
+
+export type SchemaBlocksState = {
+	validations: Record<string, BlockValidationResult>;
+};
+
+export const SchemaBlocksDefaultState: SchemaBlocksState = {
```
Xyfi, line 10: A common practice is to define this as `initialState` in the reducer itself. By returning this value from the reducer, typescript will be able to infer the type from looking at the return value of the reducer if my memory serves me right.

increddibelly: I use the default state in a selector too, so I'd prefer to keep it where both can find it

andizer (approved):
CR 👍

johannadevos:
Acceptance: 🚧 

- [x] I see validation results for the Job Posting block in the `yoast-seo/schema-blocks` store.
- [x] I see no console errors related to this PR.
- [ ] When I add other blocks, for example the Yoast Estimated Reading Time block, but also for example a Core Paragraph block, their 'validation' also appears in the store, in the `BlockValidationResult` type. Is this desired behavior?
Screenshot 2021-03-30 at 13 55 01

- [ ] The Schema Blocks validations are still present in the `yoast-seo/editor` store. I would have not expected them to be there anymore.
Screenshot 2021-03-30 at 13 56 46

- [ ] On this branch, but not on `develop` and `trunk` (in Free), the `yoast-seo/editor` and the `yoast-seo-premium/editor` store are loaded twice when adding the Job Posting block.
Screenshot 2021-03-30 at 13 58 17

PR #1123, opened 2021-03-25
Refactor the Sentence and SentencePart classes

FAMarfuaty:
## Summary

* [yoastseo] Refactors `Sentence` and `SentencePart` classes.

* `SentencePart` is changed to `Clause`.
* Since `Participle` class will be deleted in the future, the checks that are currently done inside the class are integrated to the new `Clause` class

## Test instructions

* Run yarn test and make sure all tests pass.

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/LINGO-752

### DIFF
--- a/packages/yoastseo/spec/languageProcessing/helpers/passiveVoice/periphrastic/getSentencePartsSpec.js
+++ b/packages/yoastseo/spec/languageProcessing/helpers/passiveVoice/periphrastic/getSentencePartsSpec.js
@@ -1,9 +1,9 @@
 import getSentenceParts from "../../../../../src/languageProcessing/helpers/passiveVoice/periphrastic/getSentenceParts.js";
 import arrayToRegex from "../../../../../src/languageProcessing/helpers/regex/createRegexFromArray";
-import SentencePart from "../../../../../src/values/SentencePart";
+import Clause from "../../../../../src/values/Clause";
 
 const options1 = {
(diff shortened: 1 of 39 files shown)


manuelaugustin (changes requested):
The new classes look good 👍 I have just made some small code style changes and added unit test.

However overall CR & test is 🚧 The reason for this is that on top of creating the two new classes you've also deleted the old classes and changed all the imports. I think we had talked about that during refinement, but we hadn't explicitly mentioned it on the issue. So that's a good reminder to be more explicit about the expectations next time!

I think it's better to not yet delete the old classes and keep all the imports as they were before. That way we can leave the current working structure intact for now. In your versionnow, a lot of unit tests fail because the new classes are imported in many places already where they aren't compatible.

So I think it's best to:
1) Make a new branch where you copy these new classes and their unit tests.
2) Don't delete the old classes yet (you can call the new `Sentence` class `SentenceNew` for now and we can rename it when we actually delete the old `Sentence` class).
3) Make sure all old unit tests still pass on the new branch.

FAMarfuaty:
This is the new PR https://github.com/Yoast/javascript/pull/1126 that contains only the new classes and their spec files.

iolandasequino:
Closed in favour of [#1126](https://github.com/Yoast/javascript/pull/1126)

PR #1124, opened 2021-03-26
Only output Schema when the checks for required blocks are OK

johannadevos:
## Summary

* [@yoast/schema-blocks] Fixes a bug where Schema output would still be generated if all the required blocks had been valid before, even if they were no longer valid now.
* [@yoast/schema-blocks] Fixes a bug where Schema output would be generated even when no variation had been picked for the Location block.

* If the Job Posting block overall isn't valid (`validation.result > BlockValidation.Valid`), we should dispatch a `null` Schema output to the store, to overwrite the Schema output that was potentially stored there before. 
* The Job location block template has multiple instructions that are validated. Therefore, these results should be combined in an 'overall' validation result by using `validateMany`. In order to do so, we first need to collect these results and flatten them into a list using `getAllDescendantIssues`.
* I created new values from `MissingBlock` in the `BlockValidation` enum: `MissingRecommendedBlock` and `MissingRequiredBlock`. This is because a missing recommended block should not make the overall validation result invalid, whereas a missing required block should. Now we can make that distinction.

## Test instructions

* Add a Job Posting block to your post. Don't fill in anything.
* Publish the post, inspect the Schema output, and see that the `JobPosting` graph piece is **absent**.
* Fill in the title and description, but don't choose a location in the location picker.
    * Move the focus to another block to make sure the validations are executed (should be fixed in https://yoast.atlassian.net/browse/P2-780).
    * Update the post, see that the `JobPosting` graph piece is still **absent**.
* Choose a location in the location picker (preferably the remote location, because that one is already filled in).
    * Move the focus to another block to make sure the validations are executed.
    * Update the post, see that the `JobPosting` graph piece is now **present**.
* Remove a recommended block. 
    * Move the focus to another block to make sure the validations are executed.
    * See an orange bullet for that recommended block appear in the analysis.
    * See that the analysis conclusion is still green.
    * Update the post, see that the `JobPosting` graph piece is still **present**.
* There is a known bug: if you delete the text in the remote location, the `JobPosting` graph piece will stay present but it should be absent. This is due to https://yoast.atlassian.net/browse/P2-807. 
* Empty the title and/or description (don't remove them).
    * Move the focus to another block to make sure the validations are executed.
    * Update the post, see that the `JobPosting` graph piece is now **absent**.

## Impact check

  *

## UI changes

## Quality assurance

Fixes https://yoast.atlassian.net/browse/P2-695

### DIFF
--- a/packages/schema-blocks/src/functions/presenters/SidebarWarningPresenter.ts
+++ b/packages/schema-blocks/src/functions/presenters/SidebarWarningPresenter.ts
@@ -2,8 +2,8 @@ import { getBlockType } from "../BlockHelper";
 import { select } from "@wordpress/data";
 import { __ } from "@wordpress/i18n";
 import { get } from "lodash";
-import { BlockValidationResult } from "../../core/validation";
-import { BlockValidation } from "../../core/validation";
+import { BlockValidation, BlockValidationResult } from "../../core/validation";
+import { getAllDescendantIssues } from "../validators/getAllDescendantIssues";
 
 const analysisMessageTemplates: Record<number, string> = {
 	[ BlockValidation.MissingBlock ]: "The '{child}' block is {status} but missing.",
@@ -80,21 +80,6 @@ function getAnalysisConclusion( validation: BlockValidation, issues: analysisIss
 			color: "green",
 		} as sidebarWarning;
 	}
-}
-
-/**
- * Gathers all validation issues recursively and flattens them into one list.
- *
- * @param validation The root validation result.
- *
- * @return all validation results.
- */
-function getAllDescendantIssues( validation: BlockValidationResult ): BlockValidationResult[] {
-	let results = [ validation ];
-	validation.issues.forEach( issue => {
-		results = results.concat( getAllDescendantIssues( issue ) );
-	} );
-	return results;
 }
 
 /**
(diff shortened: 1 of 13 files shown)


packages/schema-blocks/src/functions/presenters/SidebarWarningPresenter.ts
```diff
@@ -2,8 +2,8 @@ import { getBlockType } from "../BlockHelper";
 import { select } from "@wordpress/data";
 import { __ } from "@wordpress/i18n";
 import { get } from "lodash";
-import { BlockValidationResult } from "../../core/validation";
-import { BlockValidation } from "../../core/validation";
+import { BlockValidation, BlockValidationResult } from "../../core/validation";
+import { getAllDescendantIssues } from "../validators/getAllDescendantIssues";
```
increddibelly, line 6: mooier om in `validators/index.ts` deze functie te importeren en exporteren, zodat je hier
```suggestion
import { getAllDescendantIssues } from "../validators";
```
kunt doen

packages/schema-blocks/src/functions/validators/getAllDescendantIssues.ts
```diff
@@ -0,0 +1,16 @@
+import { BlockValidationResult } from "../../core/validation";
```
increddibelly, line 1: zie comment op `SidebarWarningPresenter`, dit is mooier.

increddibelly (changes requested):
even de import toevoegen aan index.ts, zie comment bij `SidebarWarningPresenter`

packages/schema-blocks/src/functions/presenters/SidebarWarningPresenter.ts
johannadevos: Done

increddibelly (approved):

andizer:
Acceptance 👍

PR #1144, opened 2021-04-06
Create three categories in the BlockValidation enum

johannadevos:
## Summary

* [@yoast/schema-blocks] Refactors the `BlockValidation` enum to group valid, okay and invalid results into three categories.

* This is a change proposed by @herregroen. The motivation is that it will be easier to write logic based on the three categories of 'validness'.
* I created a new enum value `OK`, for those validation results that should result in an orange bullet (for reasons other than being a missing recommended block, which has its own value in the enum).
* I simplified the `isValidResult` function to just check whether the result is `< 300` (because all invalid results are 300+). A change from the previous version of the code is that the `Skipped` result now is considered valid too, whereas previously it was invalid (because it was absent in `isValidResult`). I think this is not a problem, because we should not see this result anyway.

## Test instructions

* Test https://github.com/Yoast/javascript/pull/1124 again, following the test instructions there.

## Impact check

  *

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/packages/schema-blocks/src/core/validation/BlockValidation.ts
+++ b/packages/schema-blocks/src/core/validation/BlockValidation.ts
@@ -1,18 +1,28 @@
 export enum BlockValidation {
-	/** This schema block is defined to recommend a particular inner block, but that block doesn't exist. */
-	MissingRecommendedBlock = -3,
-	/** This block was skipped during validation, on purpose. If you ever see this value, that's a bug.  */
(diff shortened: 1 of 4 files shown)


johannadevos:
Closing this PR in favor of https://github.com/Yoast/javascript/pull/1145, because in the end it is unrelated to P2-789 and shouldn't close that Jira issue when it gets merged.

PR #1145, opened 2021-04-06
P2-837 Refactor block validation enum

johannadevos:
## Summary

* [@yoast/schema-blocks] Refactors the `BlockValidation` enum to group valid, okay and invalid results into three categories.

* This is a change proposed by @herregroen. The motivation is that it will be easier to write logic based on the three categories of 'validness'.
* I created a new enum value `OK`, for those validation results that should result in an orange bullet (for reasons other than being a missing recommended block, which has its own value in the enum).
* I simplified the `isValidResult` function to just check whether the result is `< 300` (because all invalid results are 300+). A change from the previous version of the code is that the `Skipped` result now is considered valid too, whereas previously it was invalid (because it was absent in `isValidResult`). I think this is not a problem, because we should not see this result anyway.

## Test instructions

* Test https://github.com/Yoast/javascript/pull/1124 again, following the test instructions there.
* Do a smoke test on the sidebar analysis; do you see the bullets you expect when making changes (for example, adding/removing required and recommended blocks, having an empty required block, having a job title that equals the post title)?

## Impact check

  *

## UI changes

## Quality assurance

Fixes #


hansjovis (approved):

hansjovis:
Acceptance test: ✅ 

* Tested https://github.com/Yoast/javascript/pull/1124 again.
* Did a smoke test on the analysis, everything still works as expected.

PR #1155, opened 2021-04-13
P2-814 Implement appender according to design

hansjovis:
## Summary

* [@yoast/schema-blocks] Fixes the styling of the block appender used in the `InnerBlocks` block instruction. 

* 

## Test instructions

* Create a post.
* Add a job posting block.
* Select the job posting block (if it is not selected already).
* You should see a block appender at the bottom of the job posting.
	* It should look like [the block appender in the design](https://www.sketch.com/s/ab6ebb5c-a390-434c-bdf1-8e3240cadcef/a/e7aAGz).
	  Screenshot 2021-04-13 at 11 22 38
	* **Note**: It may look slightly different based on the theme you have installed on your website. E.g. a different background and/or font. All in all the styling should look consistent with the rest of the theme. 
* Make sure that the block appender works as expected. 
	* E.g. remove the job description from the job posting and re-add it via the block appender.

## Impact check

  * The location and base salary inner blocks of the job posting, which have inner blocks to be able to add variations (e.g. remote location or office location), should **not** have an appender.

## UI changes

## Quality assurance

Fixes #


packages/schema-blocks/css/schema-blocks.css
```diff
@@ -50,6 +50,10 @@
 	color: #0073aa;
 }
 
+.yoast-schema-block-appender__content {
```
increddibelly, line 53: if the goal is to make a component look consistent with other components - why do we need *new* styling?

packages/schema-blocks/src/functions/presenters/BlockAppender.tsx
```diff
@@ -0,0 +1,35 @@
+import { createElement } from "@wordpress/element";
+import { Inserter } from "@wordpress/block-editor";
+
+/**
+ * Properties for the BlockAppender.
+ */
+export interface BlockAppenderProps {
```
increddibelly, line 7: not used. maybe less is more this time?

packages/schema-blocks/src/functions/presenters/BlockAppender.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Block Appender for adding an inner block to a schema block.
+ *
+ * @param clientId The id of the parent block in which a block should be added when using the appender.
+ * @param label A string
+ *
+ * @constructor
+ */
+export default function BlockAppender( { clientId, label }: BlockAppenderProps ) {
```
increddibelly, line 20:
```suggestion
export default function BlockAppender( clientId: string, label: string ) {
```

packages/schema-blocks/src/functions/presenters/BlockAppender.tsx
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Properties for the BlockAppender.
+ */
+export interface BlockAppenderProps {
+	clientId: string;
+	label: string;
+}
+
+/**
+ * Block Appender for adding an inner block to a schema block.
```
increddibelly, line 13:
```suggestion
 * Block Appender for adding a new block to an innerblocks block.
```

packages/schema-blocks/src/functions/presenters/BlockAppender.tsx
```diff
@@ -0,0 +1,35 @@
...
+		<div
+			data-root-client-id={ clientId }
+			className="wp-block block-editor-default-block-appender"
+		>
+			<p className="block-editor-default-block-appender__content yoast-schema-block-appender__content">
+				{ label }
+			</p>
+			<Inserter
+				rootClientId={ clientId }
+				position="bottom right"
```
increddibelly, line 31: bottom? is that correct?

packages/schema-blocks/src/functions/presenters/BlockAppender.tsx
hansjovis: We can implement this change, but that would mean that we can't use this function as a React component. That would also mean that we can't use JSX to instantiate this component.

[React components are functions (or classes) that get a single `props` object as its argument.](https://reactjs.org/docs/components-and-props.html#function-and-class-components)

packages/schema-blocks/src/functions/presenters/BlockAppender.tsx
hansjovis: I copied it from the `DefaultBlockAppender`. 

I guess if the space for the 'Add a job to your job posting' label is small enough that its text wraps multiple lines, we want the block appender to be at the bottom right?

packages/schema-blocks/src/functions/presenters/BlockAppender.tsx
hansjovis: See [my other comment](https://github.com/Yoast/javascript/pull/1155/files#r612361618) 😉 .

Unfortunately, if we want to use the `BlockAppender` as a React component, we need to have a single `props` argument to the function. And TypeScript expects that each argument has an appropriate type.

packages/schema-blocks/css/schema-blocks.css
hansjovis: Without this CSS the styling looks inconsistent with the other components (e.g. the 'Start writing or type / to choose a block' label). Without it, the label has 100% opacity.

But you are right that we should not ad any styling if absolutely necessary. I will look if I can add a class from WordPress so it gets the styling it needs from WordPress.

hansjovis: I looked into it. No class that I added solves the styling issue 😞.

It is really weird. If I use the browser inspector to inspect the WordPress elements from which I want to copy the styling, the inspector says that its color is `rgb(30, 30, 30)`, but that does not correspond with how the elements look inside the editor.

packages/schema-blocks/src/functions/presenters/BlockAppender.tsx
increddibelly: awesome! thanks.

packages/schema-blocks/src/functions/presenters/BlockAppender.tsx
increddibelly: that's great. thanks

packages/schema-blocks/css/schema-blocks.css
increddibelly: we're using a data-label for the label. could it be better with a `<P>` ?

increddibelly (approved):

johannadevos:
Acceptance: ✅ 

- [x] The appender becomes visible when the Job Posting block is selected.
- [x] It is possible to add a block through the appender.
- [x] The styling looks good.
- [x] The variation blocks don't have an appender.
- [ ] It is a known issue that clicking on appender selects the parent block: https://yoast.atlassian.net/browse/P2-782.

PR #1164, opened 2021-04-15
Check if block is active before activating the sidebar

andizer:
## Summary

* [@yoast/schema-blocks] Fixes a bug where the block inserter in the left sidebar crashes when hovering the yoast jobs block icon.

* I found out that hovering makes the editor crashes because it wants to activate and unexisting sidebar. I suspect that the preview is rendering the edit function from the blocktype. That function tries to activate the sidebar. Because there is no block active, it will crash. 
* Now I check if the block is selected before showing the sidebare

## Test instructions

* Create a new post.
* Click on the + icon in the top-right of the post editor to open the block insertion sidebar to the left of the editor.
* Hover over the Yoast Job Block inside this sidebar and see the editor isn't crashing
* Now insert the block and verify that the right sidebar still opens when selecting the jobs block. 
* Also navigate through the innerblocks in the jobs block to see that this also still works
* 

## Impact check

  *

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/packages/schema-blocks/src/core/blocks/BlockDefinition.tsx
+++ b/packages/schema-blocks/src/core/blocks/BlockDefinition.tsx
@@ -6,6 +6,7 @@ import {
 	BlockSaveProps,
 } from "@wordpress/blocks";
 import { InspectorControls, BlockIcon } from "@wordpress/block-editor";
+import { select } from "@wordpress/data";
 import BlockInstruction from "./BlockInstruction";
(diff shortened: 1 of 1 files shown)


johannadevos:
CR: ✅ 
I'm not testing this because I've never been able to replicate the problem on my machine, but I know that @Xyfi has.

hansjovis:
Acceptance test: ✅

PR #1182, opened 2021-04-29
Bump @wordpress/data to the latest version

johannadevos:
## Summary

* [@yoast/schema-blocks] Bumps the `@wordpress/data` dependency to its latest version.

* We were seeing failing unit tests on `release/16.3`, with the error message ` TypeError: (0 , _data.createReduxStore) is not a function`.
* This function comes from `@wordpress/data`, [see here](https://github.com/WordPress/gutenberg/tree/trunk/packages/data).
* I ran a `yarn upgrade @wordpress/data@^4.26` to bump the package to version 4.26.
* This is the version which is used [in Gutenberg 9.9](https://github.com/WordPress/gutenberg/blob/v9.9.0/packages/data/CHANGELOG.md), which is shipped with [WordPress 5.7](https://wordpress.org/news/2021/02/wordpress-5-7-release-candidate/).
* However in the end `@wordpress/data` is an external which, in production, will always match with the version shipped in Gutenberg. So we only use it in tests and for code completion.

## Test instructions

* Without this PR, see that `yarn test` in the `schema-blocks` leads to failing unit tests.
* With this PR, the tests are fixed.
* I did a quick smoke test on the Job Posting block, and saw the validation still seemed to be working (the analysis turned green after I filled in all the required blocks).

## Impact check

  * QA could do a smoke test on the Job Posting block (for example looking at the validation), but other than that this PR does not have to be tested by QA.

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/packages/schema-blocks/package.json
+++ b/packages/schema-blocks/package.json
@@ -10,7 +10,7 @@
     "@wordpress/blocks": "^6.8.0",
     "@wordpress/components": "^8.4.0",
     "@wordpress/compose": "^3.11.0",
-    "@wordpress/data": "^4.14.0",
+    "@wordpress/data": "^4.27.2",
     "@wordpress/editor": "^9.12.2",
     "@wordpress/element": "^2.9.0",
     "@wordpress/hooks": "^2.7.0",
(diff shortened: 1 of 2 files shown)


hansjovis (approved):

hansjovis:
Acceptance test: ✅ 

Unit tests pass again 🎉 .

PR #1185, opened 2021-04-30
P2-921 - Treat attribute values only containing whitespace as invalid, and number values as valid.

hansjovis:
## Summary

* [@yoast/schema-blocks] Fixes a bug where an attribute that only contains whitespace was considered valid.
* [@yoast/schema-blocks] Fixes a bug where an attribute containing a number value was considered invalid.

* Added a bunch of unit tests to test some unhappy paths.
	* Discovered a bug where number values were always considered 'empty'.
		* Decided to always say that they are valid, since a filled in number is a filled in number.  

## Test instructions

* Create a post.
* Add a job posting.
* Add whitespace (tabs, spaces, etc,), and only whitespace, to the description.
* The Schema analysis in the sidebar should give the following red result:
  > Not all required blocks have been completed! No job posting schema will be generated for your page. 
* Save the post.
* Check the Schema output of the post on the frontend.
	* It should **not** contain any `JobPosting` Schema piece. 

## Impact check

  * 

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/packages/schema-blocks/src/functions/validators/attributeNotEmpty.ts
+++ b/packages/schema-blocks/src/functions/validators/attributeNotEmpty.ts
@@ -10,7 +10,24 @@ import { isEmpty } from "lodash";
  * @returns If the attribute is considered empty.
  */
 function attributeNotEmpty( blockInstance: BlockInstance, name: string ): boolean {
-	return ! isEmpty( blockInstance.attributes[ name ] as object );
+	let value: unknown = blockInstance.attributes[ name ];
+
+	if ( typeof value === "number" ) {
+		return true;
+	}
+
+	if ( typeof value === "string" ) {
+		/*
+		 * Google trims the whitespace from any strings
+		 * (source: Google's Rich Results test tool).
+		 *
+		 * Without trimming it here as well,
+		 * values only containing whitespace would be incorrectly considered
+		 * as valid.
+		 */
+		value = value.trim();
+	}
+	return ! isEmpty( value );
 }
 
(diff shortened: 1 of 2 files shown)


increddibelly (approved):

packages/schema-blocks/src/functions/html.ts
```diff
@@ -39,6 +39,17 @@ export function stripTags( html: string, allowedTags: string[] = [] ): string {
...
 
+/**
+ * Strips all HTML tags from the string.
+ *
+ * @param html The HTML to remove the tags from.
+ *
+ * @returns The stripped HTML.
+ */
+export function stripAllTags( html: string ): string {
+	return html.replace( /(<([^>]+)>)/ig, "" );
```
johannadevos, line 50: We don't need the `i` in `/ig`, because the regex doesn't contain any case-sensitive characters.

johannadevos:
CR: ✅

johannadevos:
Acceptance: ✅ 

- [x] When entering only whitespace in the title, description or location block, the analysis stays red and no Schema output is shown on the frontend.
- [x] When adding an inline image to the title, description or location block, the analysis stays red and no Schema output is shown on the frontend.
- [x] When _also_ adding text (in addition to the inline image), the analysis turns green and the Schema output correctly shows just the text, and not the HTML for the image.
- [x] Numbers are accepted as valid input.
- [x] No console errors/warnings have been added.
- [x] It works in the default editor and with the Gutenberg plugin activated.

PR #1196, opened 2021-05-12
P2-948 - Make the Schema analysis work with blocks

hansjovis:
## Summary

* [@yoast/schema-blocks] Makes the Schema analysis work with the inner blocks now being placed directly within a post, instead of within an overarching Schema block.

* 

## Test instructions

* Companion PR of https://github.com/Yoast/wordpress-seo/pull/17022. Please test this once using the test instructions as described in that PR.

## Impact check

  *

## UI changes

## Quality assurance

Fixes #

### DIFF
--- a/packages/schema-blocks/src/functions/gutenberg/watch.ts
+++ b/packages/schema-blocks/src/functions/gutenberg/watch.ts
@@ -105,7 +105,7 @@ function clearSchemaForBlocks( block: BlockInstance ) {
 *
 * @param blocks The block instances to validate.
 *
-* @returns {BlockValidationResult[]} Validation results for each (inner)block of the given blocks.
+* @returns Validation results for each (inner)block of the given blocks.
 */
 export function validateBlocks( blocks: BlockInstance[] ): BlockValidationResult[] {
 	const validations: BlockValidationResult[] = [];
--- a/packages/schema-blocks/src/functions/gutenberg/watch.ts
+++ b/packages/schema-blocks/src/functions/gutenberg/watch.ts
@@ -1,14 +1,14 @@
 import { debounce, isEqual } from "lodash";
-import { subscribe, select, dispatch } from "@wordpress/data";
+import { dispatch, select, subscribe } from "@wordpress/data";
 import SchemaDefinition, { schemaDefinitions } from "../../core/schema/SchemaDefinition";
 import { BlockInstance } from "@wordpress/blocks";
 import warningWatcher from "./watchers/warningWatcher";
 import { getBlockDefinition } from "../../core/blocks/BlockDefinitionRepository";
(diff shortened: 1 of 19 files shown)


packages/schema-blocks/src/functions/gutenberg/watch.ts
```diff
@@ -145,7 +145,14 @@ export default function watch() {
 
 			updatingSchema = true;
 			{
-				const validations: BlockValidationResult[] = validateBlocks( rootBlocks );
+				const validationsForExistingBlocks = validateBlocks( rootBlocks );
```
johannadevos, line 148: Where did the type go?

johannadevos:
CR: ✅ 

- [x] `innerBlocksSidebar` has been renamed to `SchemaAnalysis`.
- [x] In `SchemaAnalysis`, there are no occurrences of the strings `parent` or `clientId`.
- [x] The selectors `getRequiredBlockNames` and `getRecommendedBlockNames` return a hard-coded array of block names and these arrays are complete.
- [x] `getValidationsForBlockNames` (previously `getInnerblockValidations`) has been rewritten to no longer use the parent `clientId` and directly return the array of `BlockValidationResult`s for the job posting blocks.
- [x] The sidebar is no longer rendered in the `innerBlocks` instruction.

I do see a couple of issues which will need to be split out to new tickets:
- [ ] There are a lot of traces left in the monorepo of the old Job Posting block, for example the `inheritSidebar` instruction. These will need to be removed. Many functions might need to be renamed (for example `sidebarWarningPresenter` - maybe we can just call it `warningPresenter` and leave the location unspecified).
- [ ] `useValidationResults` in `SchemaAnalysis.tsx` passes all the block names from the Yoast SEO Schema store to the `getValidationsForBlockNames` selector. This will become a problem when we have multiple structured data blocks (the analyses will become mixed up). The Schema analysis should apply to each of the (structured data blocks-related) graph pieces in turn. But I think we best leave this until the moment we add a new structured data block. There are some UX questions about this.

packages/schema-blocks/src/functions/gutenberg/watch.ts
hansjovis: TypeScript finds out automatically that `validateBlocks` returns an array of `BlockValidationResult` from the function's signature (e.g. `function validateBlocks( blocks: BlockInstance[] ): BlockValidationResult[]`). So a type is not strictly necessary.

johannadevos: Ah 💡

johannadevos:
Merging despite the failing build, because the failing unit test is in another package (`yoastseo`) and unrelated to our PR.

PR #1199, opened 2021-07-12
Update FacebookSiteAndAuthorNames.js

goaround:
Check if authorName is set and not null before check the length


IreneStr:
Hey @goaround, Thank you for your contribution! We've move to another repository, so we've ported your PR to there: https://github.com/Yoast/wordpress-seo/pull/17523

